PR #241, opened 2014-12-18
Change test tile

rafatower:
Change the tile used to test a map when first requested so that we don't
use that level of zoom nor just a corner of the raster (literaly do not
use a "corner case").

@rochoa please review


rochoa:
ping

javisantana:
yes, I'd not do this, we could be fetching million points (not likely but ok)

lbosque:
Probably because zoom 0 can be very expensive if there are million of points

rafatower:
From what I saw, using `z = 30` was failing silently in many cases so I'd prefer to remove it at all if we do not find any actual quick test that could be useful

rochoa:
Sorry but what do you mean by "using z = 30 was failing silently in many cases"? :confused:

rafatower:
Due to the bug described [here](https://github.com/mapnik/mapnik/pull/2615) (btw a failing PR, my bad) in many cases the tile requested ended up in not rendering anything because of an int overflow. So a quick test that did not test anything... or else it should have failed.

rochoa:
If the test fails the layergroup creation should fail otherwise we have a bug :-). At least we have relied on it for mapnik vectorial data until now.

rochoa:
After discussing with @rafatower and @javisantana we are gonna test different zoom levels for raster and vector data.

rafatower:
I'm having trouble with this again https://github.com/CartoDB/Windshaft/blob/277fa1a4868a84f03049541d0245dac57a001246/lib/windshaft/backends/map_validator.js#L105

rafatower:
Quite old, closing this for the moment. Let's reopen if ever needed.

PR #246, opened 2015-01-13
static preview endpoints

rochoa:
This feature includes:

 * Torque png rendering using node-canvas
 * HTTP layers
 * Static preview endpoints: being able to request a preview from a mapconfig
   * Using base layers (HTTP layers)
   * Using torque layers

Part of the functionality in server.js was split into several pieces: controllers, backends
Render cache was also split into smaller components: renderer factory and several renderers


lib/windshaft/renderers/renderer_factory.js
```diff
@@ -0,0 +1,141 @@
...
+    this.torqueRendererFactory = new TorqueRenderer.factory(); // TODO: specify options
+
+    this.blendRendererFactory = new BlendRenderer.factory(this);
+}
+
+module.exports = RendererFactory;
+
+
+RendererFactory.prototype.makeRenderer = function(mapConfig, params, callback) {
+    if ( params.layer === 'all' ) {
```
rochoa: I'm not particularly happy with this approach.

This is due to previous implementation relying on the requested format (`format` param) to decide which is the renderer instead of trying to render the requested layer (`layer` param) using the format and failing if the layer doesn't have a renderer for that format.

That would be the ideal implementation IMO. I didn't want to mix/change that behaviour during this project but most likely will change in a future iteration, so I don't consider this endpoints as official ones right now (except for the existing ones).

So ... I'm gonna add a comment about all this right here, in the code.

lib/windshaft/controllers/static_maps.js
```diff
@@ -0,0 +1,143 @@
...
+};
+
+StaticMapsController.prototype.staticMap = function(req, res, zoom, center) {
+    this._app.doCORS(res);
+
+    var renderer;
+
+    var format = req.params.format == 'jpg' ? 'jpeg' : 'png';
+
+    req.params.format = 'png';
```
rochoa: See my other comment about `format` and `layer` params and renderers.

rochoa:
Any feedback is welcome in this PR.

cc @javisantana @zenitraM @fdansv

PR #254, opened 2015-01-28
Fix grid layers interactivity order

rochoa:


rochoa:
cc @javierarce does this make sense? :smile:

PR #258, opened 2015-02-10
Mapconfig per layer datasource

rochoa:


lib/windshaft/models/mapconfig.js
```diff
@@ -40,24 +41,32 @@ function MapConfig(cfg) {
...
+        cfg: this._cfg,
+        ds: this._datasource.obj()
+    });
 };
 
 /// API: Get identifier for this MapConfig
 o.id = function() {
   if (this._id === null) {
-      this._id = this._md5(this.serialize());
+      this._id = md5Hash(JSON.stringify(this._cfg));
```
rochoa: I keep the id based on the config because if I included the datasource all tokens will change and because layers having a datasource will be always new once this is deployed so there is no need to support old configurations as they won't need a datasource.

rochoa:
@javisantana please review

More information in commit message @ https://github.com/CartoDB/Windshaft/commit/40a76e025b3172e7313f40bfc22f04708ab39df6

lib/windshaft/models/mapconfig.js
```diff
@@ -40,24 +41,32 @@ function MapConfig(cfg) {
...
 
-o._md5 = function(s) {
+function md5Hash(s) {
   return Crypto.createHash('md5').update(s).digest('hex');
 };
 
 /// API: Get serialized version of this MapConfig
 o.serialize = function() {
-  return JSON.stringify(this._cfg);
+    if (this._datasource.isEmpty()) {
```
javisantana: I understand you did this in order to not merge database auth attributes with map config and maintain backwards compatibility, right?

lib/windshaft/renderers/renderer_factory.js
```diff
@@ -88,7 +88,7 @@ RendererFactory.prototype.makeRendererTorque = function(params, mapConfig, callb
             var dbParams = RendererParams.dbParamsFromReqParams(params);
             var format = params.format;
             var layer = params.layer;
-            self.torqueRendererFactory.getRenderer(mapConfig.obj(), dbParams, format, layer, this);
+            self.torqueRendererFactory.getRenderer(mapConfig, dbParams, format, layer, this);
```
javisantana: should we remove dbParams from this now that we have that config inside the mapconfig datasource ?

javisantana:
couple of questions but looks good for me

javisantana:
:es:

lib/windshaft/models/mapconfig.js
rochoa: That's a good question! I actually did it to not store the datasource if it's empty so we keep it as small as possible in redis but it would be possible to start serialising it even it's empty because the `create` method used from MapStore will take care of the instantiation in a backwards compatible way.

lib/windshaft/renderers/renderer_factory.js
rochoa: That would be a good improvement, the thing is right now we don't store all db params (aka "per layer datasource") for all layers but just for the ones that are not using the default one to keep the previous behaviour as it was.

In next iteration we should start storing the datasource for all layers that way we could remove the dbparams parameters from here (and other dependant entities) and what is better not having to hit redis for datasource credentials when asking tiles through the layergroup id because the mapconfig would have all the required params.

I didn't do that yet because there are a lot of scenarios with /tiles/:table/* endpoints and authentication.

lib/windshaft/models/mapconfig.js
javisantana: it's ok for me

PR #262, opened 2015-02-16
Mapnik query improvements

rochoa:
Improvements for Mapnik queries using ST_SnapToGrid and ST_ClipByBox2D. New queries can be enabled via config:
 * `renderer.snapToGrid=true` will ONLY use ST_SnapToGrid
 * `renderer.snapToGrid=true` and `renderer.clipByBox2d=true` will use both: ST_SnapToGrid and ST_ClipByBox2D
   - ST_ClipByBox2D requires Postgis >=2.2
   - This requires function CDB_ScaleDenominatorToZoom


javisantana:
\o/

PR #286, opened 2015-03-02
tilelive-mapnik using fromString async method

rochoa:


package.json
```diff
@@ -32,7 +32,7 @@
         "grainstore": "~0.23.0",
         "express": "~2.5.11",
         "tilelive": "~4.5.3",
-        "tilelive-mapnik": "https://github.com/CartoDB/tilelive-mapnik/tarball/cdb",
+        "tilelive-mapnik": "https://github.com/CartoDB/tilelive-mapnik/tarball/map-from-string-async",
```
rochoa: This is a branch just to freeze the revision of tilelive-mapnik, once there is an official release we will use it from npmjs.

See: https://github.com/mapbox/tilelive-mapnik/issues/58#issuecomment-76412959

lib/windshaft/renderers/mapnik/factory.js
```diff
@@ -104,7 +104,7 @@ MapnikFactory.prototype.getRenderer = function(mapConfig, params, format, layer,
                 bufferSize: self._mapnik_opts.bufferSize,
                 scale: scaleFactor,
                 tileSize: tileSize,
-                autoLoadFonts: false,
+                autoLoadFonts: 'false',
```
rochoa: Our previous tilelive-mapnik version was using a patched version that accepted boolean values instead of 'false' string.

rochoa: See https://github.com/mapbox/tilelive-mapnik/issues/90

rochoa:
@javisantana How do you feel about start using mapnik async map creation again? :smiling_imp:

PR #289, opened 2015-03-03
Properly use bash instead of sh

christianbundy:
The Bourne shell doesn't support `[[` and throws errors when not executed under bash.

I found this bug by running `make` in the root project directory in Ubuntu 14.04 in the last released version, but I see that the shell is being set in the `Makefile` now. This no longer affects `make`, but should still be corrected.

```console
vagrant@vagrant-ubuntu-trusty-64:~/Windshaft$ ./scripts/install.sh
./scripts/install.sh: 3: ./scripts/install.sh: [[: not found
```

### DIFF
--- a/scripts/check-node-canvas.sh
+++ b/scripts/check-node-canvas.sh
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/bin/bash
 
 if [[ "$OSTYPE" == "darwin"* ]]; then
     CAIRO_PKG_CONFIG=`pkg-config cairo --cflags-only-I 2> /dev/null`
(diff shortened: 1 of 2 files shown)


rochoa:
Thanks again!! :+1: 

I did some changes so `make` workaround that (https://github.com/CartoDB/Windshaft/commit/5e638ce04b01ca8da37268441a9dfa48a265413f) but I agree it makes sense to fix it there as well! :smile:

christianbundy:
:raised_hands: Thanks for getting this merged so quickly. Cheers!

PR #297, opened 2015-03-05
Torque server side renderer using buffer-size

### LINKED ISSUE #292: Allow buffer-size in torque server side rendering
When rendering individual torque tiles in server side if one tile has to render a feature from data of an adjacent tile it won't do it because it's not considering features outside its bounding box.

rochoa:
As tiles in server are rendered individually some features that belong to a
tile in {x: 0, y: 0} but are close to the edge should be rendered in also in
tile, let's say, {x: 1, y: 0} however that last time won't have the data
because it is outside of its bounding box. Adding a bit of buffer to the
bounding box allows to render those features from others' tiles data.

Torque's windshaft provider uses by default a Uint8Array for the coordinates
so it is not possible to represent values different than 0..255 that's why we
force the use of Int16Array to be able to represent negative numbers and values
bigger than 255.

The torque renderer now accepts an options parameter that allows to setup the
buffer size for the query, by default it uses buffer-size=0. So existing users
will keep the previous behaviour.

The PngRenderer will use the provided value through Map{} for the buffer-size
or a default value of buffer-size=32.

The torque query changes to take into account the buffer size. When a value for
the buffer-size is provided the query will *pan* the values based on it. It also
takes care of the resolution to *pan* based on it.

Changes existing tests to use buffer-size=0 so they keep the old behaviour.

Closes #292

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,8 @@
-# Version 0.38.3
+# Version 0.39.0
 2015-mm-dd
+
+New features:
+ - Allow buffer-size in torque server side rendering (#292)
 
 Bugfixes:
  - Support for torque heatmaps server side rendering (#294)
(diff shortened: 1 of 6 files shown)


rochoa:
@javisantana could you take a look at this one? :smiley:

rochoa:
cc @fdansv probably you should be aware of these changes :wink:

PR #301, opened 2015-03-17
Remove trailing whitespace

christianbundy:
I've gone through these by hand and made sure that all tests pass, but it's worth mentioning that this was done automagically with:

``` shell
find . -type f \( \
  -name '*.js' -o -name '*.css' -o -name '*.md' -o -name 'README' \
\) -exec sed --in-place 's/[[:space:]]\+$//' {} \+
```

The only place where I could imagine this being a problem would be [hard line breaks in Markdown](http://spec.commonmark.org/0.17/#hard-line-break), but there aren't any of those in this repo.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -420,7 +420,7 @@ Enhancements:
  Enhancements:
 
  - Do not send 0 timings to statsd (fixes "Bad line: 0,ms" messages)
- - Do not call req2params more then once on MapConfig creation (#157) 
+ - Do not call req2params more then once on MapConfig creation (#157)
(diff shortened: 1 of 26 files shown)


rochoa:
First of all thanks for keeping your contributions coming :+1: 

I'm working on a small refactor to remove some deprecated endpoints, it will probably mean the first major release for Windshaft. Branch is https://github.com/CartoDB/Windshaft/tree/259-remove-old-api

I'm gonna hold the PR a bit because I will try today to get this into that branch to check how *conflicting* is :smile:

rochoa:
PR merged!
:+1: thanks

PR #343, opened 2015-05-22
Blend rendering allows to filter the layers to render

rochoa:
`GET :base_url_mapconfig/:TOKEN/:LAYER_FILTER/{z}/{x}/{y}.png`

`:LAYER_FILTER` supports two formats:

 - **all**: it will blend all layers in the layergroup.
 Example: `GET :base_url_mapconfig/:TOKEN/all/{z}/{x}/{y}.png`
 - **filtering by layer index** using commas: something like `0,3,4` will filter and blend layers 0,3,4.
 Example: `GET :base_url_mapconfig/:TOKEN/0,3,4/{z}/{x}/{y}.png`
 Some notes about filtering:
  * Invalid index values or out of bound indexes will end in `Invalid layer filtering` errors.
  * Once a mapnik layer is selected all mapnik layers will get blended. As this may change in the future **it is
  recommended** to always select all mapnik layers if you want to select at least one so you will get a consistent
  behaviour in the future.
  * Ordering is not considered. So right now filtering layers 0,3,4 is the very same thing as filtering 3,4,0. As this
  may change in the future **it is recommended** to always select the layers in ascending order so you will get a
  consistent behaviour in the future.

**Important**: currently format is limited to `png`. That means that all renderers in the layergroup must support that format to be able to blend them.


rochoa:
@fdansv do you mind having a look? :pray:

fdansv:
This is awesome, really looking forward to playing with the filter. Everything looks good to me, bearing in mind that at the moment I'm not that familiar with the Windshaft codebase.

:ship: :it:

rochoa:
Thanks :smile:

PR #348, opened 2015-06-01
Fixtests

BHare1985:
Fixed tests and or verified manually they pass, even if functionality of tests isn't right.


BHare1985:
Pull request to wrong repo!

PR #356, opened 2015-06-11
Checks the type of torque-specific CSS rules

### LINKED ISSUE #150: Malformed torque CartoCSS: Error: Cannot read property '0' of undefined
Using unquoted values for some torque attributes result in a weird error message.
Example unquoted value:

-torque-time-attribute:cartodb_id;

Error:

{"error":"Error: Cannot read property '0' of undefined"}

Needs an automated testcase

\cc @javisantana

fdansv:
Fixes #150 
Sending a malformed CSS style in the query will make Windshaft respond with a JSON error like the following:

```javascript
{
    errors: [
    "Unexpected type for property '-torque-aggregation-function', expected string"
    ]
}
```

@rochoa what do you think?
cc @javisantana

### DIFF
--- a/lib/windshaft/renderers/torque/factory.js
+++ b/lib/windshaft/renderers/torque/factory.js
@@ -119,6 +119,7 @@ function attrsFromCartoCSS(cartocss, required) {
         this.errors.push(e);
       }
   };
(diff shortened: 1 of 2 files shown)


rochoa:
Thanks :heart:

PR #392, opened 2015-10-07
Adds support for substitution tokens in Torque

rochoa:
Supported substitution tokens are:
 - !bbox! spatial filter applied in the tile
 - !pixel_width! width of a pixel in geographical units
 - !pixel_height! height of a pixel in geographical units
 - !scale_denominator! display pixel size

They have the same semantics than the ones in Mapnik's PostGIS plugin. For
more information about them check:
https://github.com/mapnik/mapnik/wiki/PostGIS#advanced-usage

Check https://github.com/mapnik/mapnik/wiki/ScaleAndPpi for more information
about scale_denominator.


rochoa:
What do you think @javisantana?

PR #394, opened 2015-10-28
Upgrade mapnik to 1.4.15-cdb3 and all dependants dependencies

rochoa:
I created tags in all dependencies depending on mapnik so it's easier to track what versions we are using. Previously a branch was being used so it was harder to track what was in each release / windshaft version.


rochoa:
cc @zenitraM

zenitraM:
:+1:

PR #411, opened 2015-12-24
Cuts geojson geometries into tiles to avoid a lot of redundancy in client

fdansv:
@dgaubert let's integrate this in the mapconfig+geojson branch so that we reduce the stress of the renderer right away? It's still missing things like simplifying features and so on but this change reliefs the renderer from a massive burden of redundant features.

cc @rochoa

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,7 +7,7 @@ env:
 before_install:
  - sudo add-apt-repository -y ppa:cartodb/cairo
  - sudo apt-get update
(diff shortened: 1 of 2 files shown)


dgaubert:
:+1:

rochoa:
I updated the branch with mapconfig-extension as that one has some fixes for failing tests.

About travis: I fixed a typo with libgif-dev apt package, and replaced custom ppa for python with postgresql-plpython-9.3.

PR #413, opened 2015-12-30
started to edit mapconfig content, added reviewer question for Raul

csobier:
For [Docs issue#570](https://github.com/CartoDB/docs/issues/570)

@rochoa , is is possible to break up this content by arguments/returns/examples, or is the flavor of the content different since it is a configuration file? I am trying to determine the best format to describe this content for end-users (is there another CartoDB example of configuration descriptions?)

Also, I entered a comment to you about what to do with the TODO and History sections.  Seems weird to display that to the public documentation audience.

### DIFF
--- a/doc/MapConfig-1.4.0.md
+++ b/doc/MapConfig-1.4.0.md
@@ -1,9 +1,9 @@
-# 1. Purpose
+# Purpose
 
 This specification describes [MapConfig](MapConfig-specification.md) format version 1.4.0.
 
 
-# 2. File format
+# File Format
 
 Layergroup files use the JSON format as described in [RFC 4627](http://www.ietf.org/rfc/rfc4627.txt).
 
@@ -54,7 +54,7 @@ Layergroup files use the JSON format as described in [RFC 4627](http://www.ietf.
 }
 ```
 
-## 2.1 Mapnik layers options
+## Mapnik Layers Options
 
 ```javascript
 {
@@ -135,7 +135,7 @@ Layergroup files use the JSON format as described in [RFC 4627](http://www.ietf.
 }
 ```
 
-## 2.2 Torque layers options
+## Torque Layers Options
 
 ```javascript
 {
@@ -198,7 +198,7 @@ Layergroup files use the JSON format as described in [RFC 4627](http://www.ietf.
 ```
 
 
-## 2.3 Http layers options
+## Http Layers Options
 
 ```javascript
 {
@@ -224,11 +224,14 @@ Layergroup files use the JSON format as described in [RFC 4627](http://www.ietf.
 }
 ```
 
-## 2.4 Plain layers options
+## Plain Layers Options
 
-Some notes:
- - At least one of the options, `color` or `imageUrl`, must be provided.
+**Note:**
+
+ - At least one of the options, `color` or `imageUrl`, must be provided
+
  - If both options are provided `color` will be the only one used.
+
 
 ```javascript
 {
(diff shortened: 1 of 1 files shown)


doc/MapConfig-1.4.0.md
```diff
@@ -252,11 +255,13 @@ Some notes:
...
 The document may be extended for specific uses.
 For example, Windshaft-CartoDB defines the addition of a "stat_tag" element
-in the config. See https://github.com/CartoDB/Windshaft-cartodb/wiki/MultiLayer-API
+in the config. See [https://github.com/CartoDB/Windshaft-cartodb/wiki/MultiLayer-API](https://github.com/CartoDB/Windshaft-cartodb/wiki/MultiLayer-API)
 
-Specification for how to name extensions is yet to be defined as of this version
+**Note:** Specifications for how to name extensions is yet to be defined, as of this version
 of MapConfig.
 
+{% comment %}writer note_csobier: 30-Dec-2015. I do not think the TODO and History topics below should appear in the end-user content, do you? Is there a way, to track this information somewhere more internal, like a Product Spec, or something?{% endcomment %}
```
rochoa: This is the actual internal documentation.

rochoa:
Have in mind that there are several versions of this spec, this is just one of them.

I wouldn't change the titles as they might be linked from other places and we will break those links.

rochoa:
If we want to make this part of the public CartoDB docs we should think how to include them, but please do not modify this files directly as they are specifications and we need to keep history and todo for future versions.

csobier:
Okay, agreed, we need to re-strategize how to share this content then? 

I know there are several versions so we need versioning control.  Additionally, we need to provide this MapConfig content for end-users, but not all the internal content from this page should appear.  And to match the format of the rest of the documentation, I would have to edit the format, but completely understand your point about how they are linked. So how can we use some of this internal doc for the public docs and maintain it consistently? Ideally, single-source doc content would be ideal but I know the file structure set-up is unique here.

csobier:
How about this @rochoa? I can create a public version of this content (and store locally within the Docs file structure), but every time you edit or change something, you will have to add a Docs issue so that I am notified to update the public docs. Does that work for you?

rochoa:
Previous versions should not be modified, if we need to add some new behaviour we need to create a new version. That's why we have https://github.com/CartoDB/Windshaft/blob/46c56495073adcd56016f4d85124a8a1c5d50a67/doc/MapConfig-specification.md pointing to latest version.

I really want to keep history and todo in these documents. As I said, this is internal documentation and the actual Maps API doesn't follow this very same spec but an extended version from Windshaft-cartodb as stated in https://github.com/CartoDB/Windshaft-cartodb/blob/45acdd9f39b8fb4aced43dda1e8a1e1c1cf85da1/docs/MultiLayer-API.md. It also adds another extension: https://github.com/CartoDB/Windshaft-cartodb/blob/45acdd9f39b8fb4aced43dda1e8a1e1c1cf85da1/docs/MapConfig-NamedMaps-extension.md which should be documented in the final/end-user documentation.

In conclusion if we want to make this public/end-user friendly we should do it with the rest of the documentation in Windshaft-cartodb.

csobier:
I'm going to be proactive and delete this PR and related [Docs repo PR#614](https://github.com/CartoDB/docs/pull/614). As we are not going to edit Windshaft content directly. See [Docs PR#622](https://github.com/CartoDB/docs/pull/622) as the suggested alternative.

PR #419, opened 2016-01-20
Overviews work

jgoizueta:
This branch changes the tiler to use overviews metadata if available in MapConfig to alter the SQL queries and make use of overviews for rendering


jgoizueta:
@rochoa @rafatower can you CR this?

Some test should be added for the MapnikFactory handling mapconfigs with overviews, though, but I think this is ready for CRing...

lib/windshaft/renderers/mapnik/factory.js
```diff
@@ -227,10 +228,14 @@ MapnikFactory.prototype.mapConfigToMMLBuilderConfig = function(mapConfig) {
 
         var lyropt = lyr.options;
 
-        style.push(lyropt.cartocss);
+        var overviews_metadata = lyropt.overviews;
+        var query = Overviews.query(lyropt.sql, overviews_metadata);
+        var cartocss = Overviews.style(lyropt.cartocss, lyropt.cartocss_version, overviews_metadata);
```
rafatower: I thought that making the `cartocss` dependent on the overviews was deferred to a later version, but ok.

lib/windshaft/utils/overviews.js
```diff
@@ -0,0 +1,146 @@
...
+    for ( var table in overviews ) {
+        if (overviews.hasOwnProperty(table)) {
+            var table_view = overviews_view_name(table);
+            cartocss = replace_table_in_style(cartocss, table, table_view);
+        }
+    }
+    return cartocss;
+}
+
+module.exports = { query: overviews_query, style: overviews_style };
```
rafatower: In other places I've seen this pattern:
```js
var MyModuleName = {
  my_function: function(args) {
     ...
  }
  my_function2: function(args) {
     ...
  }
}
module.exports = MyModuleName;
```
(I guess this avoids polluting the namespace with global functions but I'm a newbie here, @rochoa will have a more informed opinion here)

lib/windshaft/utils/table_name_parser.js
```diff
@@ -0,0 +1,86 @@
...
+        } // else invalid table name (missing dot)
+    } else if ( parts.length === 3 && parts[1].part === '.' ) {
+        schema_part = parts[0].part;
+        table_part = parts[2].part;
+    } // else invalid table name
+    return {
+      schema: schema_part,
+      table:  table_part
+    };
+}
```
rafatower: Wow, pretty impressive :D (also looked at the tests covering all this)

OTOH This is the kind of detail I'd deferred for a later iteration.

rafatower:
Overall looks good to me.

lib/windshaft/renderers/mapnik/factory.js
jgoizueta: Yes, I removed the changes being done to the CartoCSS but left the calls as a placeholder

lib/windshaft/utils/overviews.js
jgoizueta: Coming from CoffeScript I didn't expect any function defined there to be global :disappointed:
If they are global (please confirm, @rochoa?) I will definitely use that pattern.

jgoizueta: I've done some testing and it seems to me functions defined at top level of a node module don't pollute the global namespace. Can get confirmation on this?

rochoa: That's not a problem. They are not polluting the global namespace.

lib/windshaft/utils/overviews.js
```diff
@@ -0,0 +1,146 @@
...
+// Vector Overviews support functions
+
+// TODO: some names are introudced in the queries, and the
+// '_vovw_' (for vector overviews) is used in them, but no check
+// is performed for conflicts with existing identifiers in the query.
+
+// Build UNION expression to replace table, using overviews metadata
+// overviews metadata: { 1: 'table_ov1', ... }
+// assume table and overview names include schema if necessary and are quoted as needed
+function overviews_view_for_table(table, overviews_metadata, indent) {
```
rochoa: I though the initial approach was targeting simple queries like `select * from table`.

lib/windshaft/utils/overviews.js
```diff
@@ -0,0 +1,146 @@
...
+    // jshint maxcomplexity:7
```
rochoa: :smiling_imp: Refactor required, modifying jshint rules only for existing code. New code should adhere to the rules.

lib/windshaft/utils/overviews.js
```diff
@@ -0,0 +1,146 @@
...
+    for (i = 0, len = sorted_overviews.length; i < len; i++) {
+        z_hi = parseInt(sorted_overviews[i][0]);
+        ov_table = sorted_overviews[i][1];
+        overview_layers.push([overview_z_condition(z_lo, z_hi), ov_table]);
+        z_lo = z_hi;
+    }
+    overview_layers.push(["_vovw_z > " + z_lo, table]);
+
+    selects = [];
+    for (i = 0, len = overview_layers.length; i < len; i++) {
```
rochoa: I would use `selects = overview_layers.map()`.

lib/windshaft/utils/overviews.js
```diff
@@ -0,0 +1,146 @@
...
+        condition = overview_layers[i][0];
+        ov_table = TableNameParser.parse(overview_layers[i][1]);
+        ov_table.schema = ov_table.schema || parsed_table.schema;
+        var ov_identifier = TableNameParser.table_identifier(ov_table);
+        selects.push(indent + "SELECT * FROM " + ov_identifier + ", _vovw_scale WHERE " + condition);
```
rochoa: Consider using templates for string concatenation. Example: https://github.com/CartoDB/Windshaft/blob/ce4fb8dcd5f205b23d9d6b92a79f4bdced731e76/lib/windshaft/models/widgets/formula.js#L10

lib/windshaft/utils/overviews.js
```diff
@@ -0,0 +1,146 @@
...
+    // style may contain #old_table references, which have no actual effect
+    // so we leave it alone...
+    // jshint unused:false
+    return style;
+}
+
+// Substitute tables by overviews query
+// overviews contains metadata about the overviews to be used:
+//     { 'table-name': {1: { table: 'overview-table-1' }, ... }, ... }
+function overviews_query(query, overviews) {
```
rochoa: In general I find this approach hard to follow, and to maintain.

I am also worried about all the string manipulation we are doing to replace table names with overview names.

lib/windshaft/utils/table_name_parser.js
```diff
@@ -0,0 +1,86 @@
...
+        return '"' + txt.replace(/\"/g, '""') + '"';
+    } else {
+        return txt;
+    }
+}
+
+// Parse PostgreSQL table name (possibly quoted and with optional schema).+
+// Returns { schema: 'schema_name', table: 'table_name' }
+function parse_table_name(table) {
+    // jshint maxcomplexity:15
```
rochoa: This needs to be refactored to adhere to current rules.

rochoa:
In general I'm a bit worried about the complexity of the code.

I also would like to see some acceptance tests using overviews in the mapconfig.

Also, will this work with a query like:
```sql
select populated_places_simple.* from populated_places_simple join world_borders
on st_intersects(populated_places_simple.the_geom, world_borders.the_geom)
where world_borders.name = 'France'
```

I know this is not an easy problem at all, and the work put on this is pretty amazing but I'm worried and I'm starting to think if we should go with a simpler approach for now.

I mean: if you have simple queries like:
 1. `select * from table`
 1. `select cartodb_id, the_geom_webmercator from table`

You can do an initial query with limit 0 to get the column names and the prepare a query using the overviews:

1.

```sql
WITH
  _user_query AS (SELECT * FROM table),
  _vovw_scale AS ( SELECT 1 AS _vovw_z ),
  _vovw_table AS (
    SELECT cartodb_id, the_geom_webmercator, the_geom, other_table_column FROM table_ov08, _vovw_scale WHERE _vovw_z <= 8
    UNION ALL
    SELECT cartodb_id, the_geom_webmercator, the_geom, other_table_column FROM _user_query, _vovw_scale WHERE _vovw_z > 8
  )
  SELECT * FROM _vovw_table
```

2.

```sql
WITH
  _user_query AS (SELECT cartodb_id, the_geom_webmercator FROM table),
  _vovw_scale AS ( SELECT 1 AS _vovw_z ),
  _vovw_table AS (
    SELECT cartodb_id, the_geom_webmercator FROM table_ov08, _vovw_scale WHERE _vovw_z <= 8
    UNION ALL
    SELECT cartodb_id, the_geom_webmercator FROM _user_query, _vovw_scale WHERE _vovw_z > 8
  )
  SELECT * FROM _vovw_table
```

If we still want to go with complex queries and query replacements I would start considering the use of Postgresql parser, as per https://github.com/lfittl/libpg_query.

lib/windshaft/utils/overviews.js
jgoizueta: Grrrrr, it was so easy to fix everything with `// jshint` ...

lib/windshaft/utils/overviews.js
jgoizueta: It was. Along the way I thought it would be just as easy to be more flexible. We'll discuss it as for your comments below.

lib/windshaft/utils/overviews.js
jgoizueta: Right.

lib/windshaft/utils/table_name_parser.js
jgoizueta: OK (grrr) done, hope it's more clear now

lib/windshaft/utils/overviews.js
jgoizueta: Can handle iteration/multiple elements with dot? (will need to for the overview layers). I'll take a look.

jgoizueta:
I agree some acceptance tests are needed, I meant to work on that but have put it off at the moment.

Regarding your proposed simplified approach, as I see it, its main advantage is that we eliminate the string substitutions for table names, that could be a source of problems. I'm not sure the code would become significantly more simple, am I missing anything?

Maybe we could leave the current complex queries (supporting more general input queries) in a branch for further testing and implement your approach at the moment, let's discuss it.

jgoizueta:
This needs to be reviewed @rochoa @rafatower can you take a look?

I've added an acceptance test and also have make some relevant refactoring so that the detail of using `CDB_ZooomFromScale` is not defined here (but in Windshaft-cartodb). This also opens the door of providing different ways for constructing the overviews queries outside from Windshaft.

jgoizueta:
Here are the modifications needed in Windshaft-cartodb:
https://github.com/CartoDB/Windshaft-cartodb/commit/8348f745134704ef99d6af169f79d39d96aeabab

jgoizueta:
Maybe the implementation of OverviewsHandler could be moved completely out of here (to WIndshaft-cartodb). Then we would perform there the kind of acceptance tests done here and replace theses by something simpler using a mockup OverviewsHander.

lib/windshaft/index.js
```diff
@@ -30,5 +30,6 @@ module.exports = {
         grainstore: grainstore.version(),
         node_mapnik: mapnik.version,
         mapnik: mapnik.versions.mapnik
-    }
+    },
+    OverviewsHandler: require('./utils/overviews_handler')
```
rochoa: Why do we need to expose this?

rochoa: Oh, I see. To expose it for Windshaft-cartodb. I find this weird.

Either, as you were proposing, we move everything to Windshaft-cartodb or Windshaft-cartodb only adds overviews metadata.

I'm starting to be more inclined to move everything to Windshaft-cartodb so Windshaft keeps doing just one thing, rendering a SQL query.

jgoizueta:
Now this should be ready for release, @rochoa can you CR/approve?
Note: Windshaft-cartodb is adapted for this here: https://github.com/CartoDB/Windshaft/pull/419

doc/MapConfig-1.5.0.md
```diff
@@ -143,6 +143,16 @@ Layergroup files use the JSON format as described in [RFC 4627](http://www.ietf.
...
+    },
+
+    // OPTIONAL
+    // A query-rewriter can be supplied in the Windshaft server configuration
+    // to preprocess SQL queries and cartocss
+    // code before it is passed to the Mapnik renderer.
+    // Per-layer data can be passed to the query rewriter with
+    // this option:
+    query_rewrite_data: {
+      // ... arbitrary date used by the query rewriter
```
rafatower: typo: s/date/data

lib/windshaft/renderers/mapnik/factory.js
```diff
@@ -91,7 +92,18 @@ function MapnikFactory(options) {
...
+        // A query-rewriter can be passed to preprocess SQL queries and
+        // CartoCSS before it is passed to Mapnik.
+        // The query rewriter should contain to functions:
+        // - `query(sql, data)` to transform queries using the optional data provided
+        // - `style(cartocss, cartocssversion, data)` to transform cartocss
+        // The data passed to these functions can be provided for eaach layer
+        // through a `query_rewrite_data` entry in the layer options.
+        // By default a dummy query rewriter which doesn't alter queries
+        // nor style is used.
+        queryRewriter: new QueryRewriter()
```
rafatower: Shouldn't it be called DummyQueryRewriter or DefaultQueryRewriter?

lib/windshaft/utils/query_rewriter.js
```diff
@@ -0,0 +1,23 @@
+// Dummy query-rewriter which doesn't alter queries
+
+// This class implements the Windshaft query rewriting API:
+//
+// * query(sql, data) // transform SQL query, with additional data
+// * style(cartocss, casrtocss_version, data) // transform cartoCSS
+//
```
rafatower: It just feels a bit weird that the QueryRewriter rewrites queries and style. I understand they share the data, which is a convenience. I cannot come up with a better name.

On one hand, I'd split it into QueryRewriter and StyleRewriter, following the SRP (as they can still share `data`. OTOH I'm wondering if we're adding the cartocss part too soon.

doc/MapConfig-1.5.0.md
jgoizueta: thanks

lib/windshaft/renderers/mapnik/factory.js
jgoizueta: Yes, it would be more clear, I'll change it

lib/windshaft/utils/query_rewriter.js
jgoizueta: Maybe should remove the style part for the time being. It is there because initially I was also changing references to the tables in the CartoCSS, but then it seem unnecessary and now it does nothing with the CartoCSS. I vote for removing it, since when/if we actually need it we may find unexpected requierements for it...

lib/windshaft/renderers/mapnik/factory.js
```diff
@@ -91,7 +92,16 @@ function MapnikFactory(options) {
...
+        },
+
+        // A query-rewriter can be passed to preprocess SQL queries
+        // before passing them to Mapnik.
+        // The query rewriter should contain one function:
+        // - `query(sql, data)` to transform queries using the optional data provided
+        // The data passed to this function can be provided for eaach layer
+        // through a `query_rewrite_data` entry in the layer options.
+        // By default a dummy query rewriter which doesn't alter queries is used.
+        queryRewriter: new DefaultQueryRewriter()
```
rochoa: I though the plan was to pass a QueryRewriter each time we want to create a mapnik layer. But as I understand this is a configuration QueryRewriter. The reason I don't like this approach is that we need to do the query_rewrite_data mapconfig hack in order to share data with the QueryRewriter.

IMO we should pass a QueryRewriter each time we want to create a layer, and that QueryRewriter instance should have everything it needs, i.e. query_rewrite_data, bounded within the instance. That way we don't need the query_rewrite_data hack.

lib/windshaft/utils/default_query_rewriter.js
```diff
@@ -0,0 +1,16 @@
...
+// This class implements the Windshaft query rewriting API:
+//
+// * query(sql, data) // transform SQL query, with additional data
+//
+function DefaultQueryRewriter() {
+}
+
+module.exports = DefaultQueryRewriter;
+
+DefaultQueryRewriter.prototype.query = function(query, data) {
```
rochoa: See my previous comment: for me the signature should be `function(query)`.

I'm not sure if this being a public interface it should be async like `function(query, callback)`.

test/unit/default_query_rewriter.js
```diff
@@ -0,0 +1,47 @@
+require('../support/test_helper.js');
+
+var assert                 = require('assert');
+var DefaultQueryRewriter = require('../../lib/windshaft/utils/default_query_rewriter');
+var queryRewriter       = new DefaultQueryRewriter();
+
+describe('Default QueryRewriter', function() {
+
+  it('does not alter queries ', function(done){
```
rochoa: I would do an array with queries and .forEach it.

test/acceptance/query_rewriter.js
```diff
@@ -0,0 +1,60 @@
...
+var rendererOptions = global.environment.renderer;
+
+var testQueryRewriter = {
+    query: function(query, data) {
+        var table = (data && data.table) || 'test_table';
+        return 'SELECT * FROM ' + table;
+    },
+};
+
+describe('server_gettile', function() {
```
rochoa: Can we achieve the same without creating a new test table?

lib/windshaft/renderers/mapnik/factory.js
jgoizueta: I agree that would be tidier. I guess I still had in mind that the overviews metadata could have other uses but since it is inside specific "query_rewriter_data" that doesn't make much sense...

OK, I think it will be simpler that way.

lib/windshaft/utils/default_query_rewriter.js
jgoizueta: Making it asynchronous would give the user more flexibility over what's put there, obviously...

I think it won't be much trouble to make it async, we had at one point an async operation where this is used.

test/unit/default_query_rewriter.js
jgoizueta: Yes, nicer, thnx

test/acceptance/query_rewriter.js
jgoizueta: A new 'overview table' `_vovw_12_test_table` (with two points out of the existing `test_table`) has been added to the fixtures so we can detect when is the original table used for painting tiles and when is the overview table used (by comparing to reference tiles). 

We are testing here if a query rewriter is used as intended for rendering tiles, and need two different tables with noticeable differences in the same area... We could use the existing `test_big_poly` instead I guess.

rochoa: ok, fair enough! :smile:

PR #420, opened 2016-02-02
Metadata refactor

dgaubert:
cc/ @rochoa

### DIFF
--- a/lib/windshaft/metadata/empty-layer-metadata.js
+++ b/lib/windshaft/metadata/empty-layer-metadata.js
@@ -1,0 +1,19 @@
+'use strict';
+
+function EmptyLayerMetadata() {
+    this._types = ['http'];
+}
+
+EmptyLayerMetadata.prototype.is = function (type) {
+    return this._types.indexOf(type) !== -1;
+};
+
+EmptyLayerMetadata.prototype.getMetadata = function (mapConfig, layer, layerId, params, rendererCache, callback) {
+    process.nextTick(function() {
+        callback(null, {
+            stats: []
+        });
+    });
+};
+
+module.exports = EmptyLayerMetadata;
--- a/lib/windshaft/metadata/empty-layer-metadata.js
+++ b/lib/windshaft/metadata/empty-layer-metadata.js
@@ -1,11 +1,11 @@
 'use strict';
 
 function EmptyLayerMetadata(types) {
-    this._types = types || [];
+    this._types = types || {};
 }
 
 EmptyLayerMetadata.prototype.is = function (type) {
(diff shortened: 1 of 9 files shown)


lib/windshaft/metadata/index.js
```diff
@@ -0,0 +1,21 @@
+'use strict';
+
+var LayerMetadata = require('./layer-metadata');
+var EmptyLayerMetadata = require('./empty-layer-metadata');
+var MapnikLayerMetadata = require('./mapnik-layer-metadata');
+var TorqueLayerMetadata = require('./torque-layer-metadata');
+
+module.exports = function LayerMetadataFactory() {
+    var layerMetadataIterator = [];
+    var layerMetadataEnabled = global.environment.enabledFeatures.layerMetadata;
```
rochoa: Have in mind that previous versions won't have `global.environment.enabledFeatures` entry config.

lib/windshaft/metadata/index.js
```diff
@@ -0,0 +1,21 @@
...
+
+    if (layerMetadataEnabled) {
```
rochoa: I'm missing `plain` layers.

lib/windshaft/metadata/empty-layer-metadata.js
```diff
@@ -0,0 +1,19 @@
+'use strict';
+
+function EmptyLayerMetadata(types) {
+    this._types = types || [];
+}
+
+EmptyLayerMetadata.prototype.is = function (type) {
+    return this._types.indexOf(type) !== -1;
```
rochoa: µoptimization: what about using a dict/map?

lib/windshaft/metadata/index.js
dgaubert: it should have empty behaviour, Shouldn't it?

lib/windshaft/metadata/index.js
dgaubert: For previous versions, should it be enabled?

rochoa:
Nice! :shipit:

lib/windshaft/metadata/empty-layer-metadata.js
dgaubert: In es5 I must use a plain object for that, I don't know if it's appreciable.. but, of course, it can be done with a dict/map ;)

dgaubert: In es6 we have Map/WeakMap and Set/WeakSet ... but for now... plain objects..

rochoa:
:+1:

PR #430, opened 2016-02-10
Geojson buffer

fdansv:
@dgaubert @rochoa 

As discussed with @javisantana, we're not specifying dynamically buffer size for now, so it's currently a constant set to 32px.


rochoa:
Can you fix the tests? :pray:

PR #438, opened 2016-02-19
Turbocartocss integration

dgaubert:
cc @rochoa


lib/windshaft/style/postgres-datasource.js
```diff
@@ -0,0 +1,49 @@
...
+    this.query = query;
+}
+
+PostgresDatasource.prototype.getName = function () {
+    return 'PostgresDatasource';
+};
+
+PostgresDatasource.prototype.getRamp = function (column, method, callback) {
+    var methods = {
+        quantiles: 'CDB_QuantileBins(array_agg(distinct({{column}}::numeric)), 5) as quantiles',
```
rochoa: We will have to evaluate using postgresql/postgis stats for some of these methods but I consider this fair enough for now.

package.json
```diff
@@ -20,25 +20,26 @@
         "Raul Ochoa <rochoa@cartodb.com>"
     ],
     "dependencies": {
-        "underscore": "~1.6.0",
```
rochoa: :cry: I will reorder deps in master so it's more clear.

lib/windshaft/style/postgres-datasource.js
```diff
@@ -0,0 +1,50 @@
...
+function createTemplate(method) {
+    return dot.template([
+        'SELECT',
+        method,
+        'FROM ({{=it._sql}}) _table_sql WHERE {{=it._column}} IS NOT NULL'
+    ].join('\n'));
+}
+
+var methods = {
+    quantiles: 'CDB_QuantileBins(array_agg(distinct({{=it._column}}::numeric)), 5) as quantiles',
```
rochoa: I'm adding again this: we will have to reevaluate these queries as they won't be good enough is some cases.

lib/windshaft/renderers/mapnik/factory.js
```diff
@@ -196,7 +197,22 @@ MapnikFactory.prototype.getMapnikRenderer = function(mapConfig, format, options,
     }
 
     step(
-        function initBuilder() {
+        function parseCartoCss () {
+            var next = this;
+
+            var requestDbParams = RendererParams.dbParamsFromReqParams(options.params);
```
rochoa: I'm starting to see this pattern/code too often lately. We need to think about how to avoid duplicating this.

Issue #439, opened 2016-02-23
[vector] Tiles are still being cut in tile edges


fdansv:
As reported in https://github.com/CartoDB/d3.cartodb/issues/111

This has to do with a wrong calculation of the buffer distance in metres.

PR #440, opened 2016-02-23
Passes buffer size calculation to SQL query

fdansv:
Fixes https://github.com/CartoDB/Windshaft/issues/439

@javisantana


javisantana:
Looks good to me but I don't know if windshaft has cartodb-extension as requirement cc @rochoa

fdansv:
We were already using `CDB_XYZ_Extent` so I'd assume so. Merging, as discussed offline!

javisantana:
:es:

rochoa:
We will have to do this properly because we are relying on cartodb-postgresql extension in order to get this quicker but we should stick to postgis functions and calculate extent and buffer as we do for torque.

PR #451, opened 2016-03-01
Simplify geojson

### LINKED ISSUE #446: Use simplify strategy in geojson-renderer.
We should return a **simplified** version of the geometries instead a full-quality ones in order to improve the throughput. We should apply the same strategy we use in _mapnik-renderer_.

dgaubert:
Fix #446 

cc @rochoa


rochoa:
Now there are real tests failing after making the travis build to start again.

lib/windshaft/renderers/mapnik/geojson_template.sql
```diff
@@ -1,9 +1,66 @@
-SELECT row_to_json(featcoll) as geojson FROM
-  (SELECT 'FeatureCollection' AS TYPE, array_to_json(array_agg(feat)) AS features FROM
-    (SELECT 'Feature' AS TYPE, ST_AsGeoJSON({{ if (it.clipByBox2d) { }}ST_ClipByBox2D{{ } else { }}ST_Intersection{{ } }}(st_makevalid(tbl.{{= it.geomColumn }}), st_expand(CDB_XYZ_Extent({{= it.coord.x }}, {{= it.coord.y }}, {{= it.zoom }}), cdb_xyz_resolution({{= it.zoom }}) * {{= it.bufferSize}})))::json AS geometry, {{ if (!it.columns || it.columns.length === 0) { }} '{}'::json {{ } else { }} row_to_json(
-      (SELECT l FROM
-        (SELECT {{ for (var i = 0; i < it.columns.length; i++) { }} {{= it.columns[i] }} {{ if (i < (it.columns.length - 1)) { }}, {{ } }} {{ } }}) AS l
+SELECT
```
rochoa: Super-hard to find what changed here :cry:. Could we change the formatting in another PR, merge it to master, and then we can compare this easier. :pray:

lib/windshaft/renderers/mapnik/geojson_template.sql
```diff
@@ -1,9 +1,10 @@
...
+    (SELECT 'Feature' AS TYPE, st_asgeojson( {{ if (it.clipByBox2d) { }}st_clipbybox2d({{ } else { }}st_intersection({{ } }}st_makevalid(st_simplify({{ if (it.removeRepeatedPoints) { }}st_removerepeatedpoints(st_makevalid(tbl.{{= it.geomColumn }}), cdb_xyz_resolution({{= it.zoom }}) * (1.0 / 20.0)), {{ } else { }}st_makevalid(tbl.{{= it.geomColumn }}), {{ } }}cdb_xyz_resolution({{= it.zoom }}) * (1.0 / 20.0))), st_expand(cdb_xyz_extent({{= it.coord.x }}, {{= it.coord.y }}, {{= it.zoom }}), cdb_xyz_resolution({{= it.zoom }}) * {{= it.bufferSize}}))
+      )::json AS geometry, {{ if (!it.columns || it.columns.length === 0) { }} '{}'::json {{ } else { }} row_to_json(
+        (SELECT l FROM
+          (SELECT {{ for (var i = 0; i < it.columns.length; i++) { }} {{= it.columns[i] }}{{ if (i < (it.columns.length - 1)) { }}, {{ } }} {{ } }}) AS l
+        )
+      ) {{ } }} AS properties
+      FROM ({{= it.layerSql }}) AS tbl
+      WHERE (st_intersects({{= it.geomColumn }}, st_expand(cdb_xyz_extent({{= it.coord.x }}, {{= it.coord.y }}, {{= it.zoom }}), cdb_xyz_resolution({{= it.zoom }}) * {{= it.bufferSize}})))) AS feature) AS featurecollection;
```
dgaubert: It wasn't as I expected but I think is better.

rochoa: It was much better the other way,  wasn't it?

What I meant was to reformat the code in master and the change it here so we can spot what have changed.

rochoa: I'm gonna try to do that :smile:

rochoa:
I guess not it's a bit easier to understand the changes :smile:

lib/windshaft/renderers/mapnik/geojson_template.sql
```diff
@@ -19,12 +28,13 @@ FROM (
                 row_to_json((SELECT l FROM (SELECT {{=it.columns}}) AS l))
             {{ } }} AS properties
             FROM ({{= it.layerSql }}) AS tbl
-            WHERE
-                {{= it.geomColumn }} &&
-                    st_expand(
-                        CDB_XYZ_Extent({{= it.coord.x }}, {{= it.coord.y }}, {{= it.zoom }}),
-                        cdb_xyz_resolution({{= it.zoom }}) * {{= it.bufferSize}}
-                    )
+            WHERE st_intersects(
```
rochoa: Why did we move from `&&` to `st_intersects`?

lib/windshaft/renderers/mapnik/geojson_template.sql
```diff
@@ -6,7 +6,16 @@ FROM (
             SELECT 'Feature' AS TYPE,
             ST_AsGeoJSON({{=it.clipFn}}(
                 st_makevalid(
```
rochoa: Should we remove this `st_makevalid` as we are using it inside `st_simplify` also?

lib/windshaft/renderers/mapnik/geojson_template.sql
rochoa: cc @pnorman

lib/windshaft/renderers/mapnik/geojson_template.sql
```diff
@@ -6,7 +6,16 @@ FROM (
             SELECT 'Feature' AS TYPE,
             ST_AsGeoJSON({{=it.clipFn}}(
```
rochoa: I would like to go with the same approach we have in mapnik postgis_datasource where we start with the functions that drop more information sooner: https://github.com/CartoDB/mapnik/blob/8cf77fa91467e31b6250861bfa0f23f9d622e055/plugins/input/postgis/postgis_datasource.cpp#L762-L796

So clipping should be the first function to apply :smiling_imp:.

lib/windshaft/renderers/mapnik/geojson_template.sql
```diff
@@ -6,7 +6,16 @@ FROM (
...
                 st_makevalid(
-                    tbl.{{= it.geomColumn }}
+                    st_simplify(
+                        {{ if (it.removeRepeatedPoints) { }}
+                            st_removerepeatedpoints(
+                                st_makevalid(tbl.{{= it.geomColumn }}
+                            ), cdb_xyz_resolution({{= it.zoom }}) * (1.0 / 20.0)),
```
rochoa: And last but not least important we cannot rely on cdb_* functions, so we need to calculate the envelopes and buffers without relying on these functions.

lib/windshaft/renderers/mapnik/geojson_template.sql
pnorman: Some of the PostGIS calls in the column list are expensive, and reducing the number of geometries by using ST_Intersects dramatically speeds up the query.

rochoa: Another point after talking to @pnorman: this also helps to discard empty geometries when we clip the geometry. That problem could be solved also by discarding null geometries after doing all simplifications/clippings.

lib/windshaft/renderers/mapnik/geojson_template.sql
rochoa: See https://github.com/CartoDB/Windshaft/blob/93fb256f211744e4e5f102046d867f1d337c8981/lib/windshaft/renderers/torque/renderer.js#L66-L159

lib/windshaft/renderers/mapnik/geojson_template.sql
```diff
@@ -1,30 +1,38 @@
...
             SELECT 'Feature' AS TYPE,
-            ST_AsGeoJSON({{=it.clipFn}}(
-                st_makevalid(
-                    tbl.{{= it.geomColumn }}
-                ),
-                st_expand(
-                    CDB_XYZ_Extent({{= it.coord.x }}, {{= it.coord.y }}, {{= it.zoom }}),
-                    cdb_xyz_resolution({{= it.zoom }}) * {{= it.bufferSize}}
+            ST_AsGeoJSON(
+                ST_Simplify(
```
rochoa: The problem I see with this approach is that we will include empty/null geometries because of the simplification so we will still transfer features that we are not interested due to the simplification.

I'm gonna try to add an acceptance test to cover that so we can see better what I mean.

lib/windshaft/renderers/mapnik/geojson_template.sql
```diff
@@ -1,30 +1,39 @@
...
+                        ST_MakeEnvelope({{= it.extent.xmin }}, {{= it.extent.ymin }}, {{= it.extent.xmax }}, {{= it.extent.ymax }}, {{= it.srid }}),
+                        {{= it.xyzResolution }} * {{= it.bufferSize}}
                     )
-        ) AS feat
-    )
-AS featcoll;
+                )
+            )
+        ) AS feature
+        WHERE geometry IS NOT NULL
```
rochoa: I already added this to skip null geometries but still we will send empty coordinates.

lib/windshaft/renderers/mapnik/geojson_template.sql
rochoa: See #464 should fix this problem.

rochoa:
cc @fdansv beware of commit https://github.com/CartoDB/Windshaft/pull/451/commits/7a8de475c8077150b91de3eff8b4dc8283deae2f

fdansv:
> cc @fdansv beware of commit 7a8de47

Sounds good!

rochoa:
@fdansv btw, that should be covered by https://github.com/CartoDB/d3.cartodb/pull/152

rochoa:
What do you think about current state @dgaubert? Good enough?

dgaubert:
Well, I've been checking the final query and I really like the way of (dump-simplify-collect)*ing geometries. So, I'd say yes, it is good enough.

:+1:

rochoa:
I'm a bit worried about the performance of that part when there are complex multigeometries and dumping/collecting them will get slow and use a lot of CPU at database. cc @javisantana

dgaubert:
I'd try to see how it works, and then, make a decision

PR #455, opened 2016-03-03
Use pg 9.3 in travis builds

rochoa:


rochoa:
No longer required, replaced by #456

Issue #453, opened 2016-03-03
Move turbo-cartocss to windshaft-cartodb


rochoa:
Windshaft responsibility is to take a cartocss and render with it. It should not be the one responsible for preprocessing cartocss.

dgaubert:
Fixed in #459

PR #457, opened 2016-03-04
Passes cartocss in metadata

### LINKED ISSUE #452: CartoCSS should be sent in metadata
We need this so that d3.cartodb is able to style maps

fdansv:
Fixes #452 

cc @dgaubert

### DIFF
--- a/lib/windshaft/metadata/mapnik-layer-metadata.js
+++ b/lib/windshaft/metadata/mapnik-layer-metadata.js
@@ -80,6 +80,7 @@ MapnikLayerMetadata.prototype.getMetadata = function (mapConfig, layer, layerId,
             }
 
             callback(null, {
+                cartocss: layer.options.cartocss,
                 stats: result
             });
         }


rochoa:
Should we add this also to torque layers?

rochoa:
One extra question, is this returning the processed cartocss (after applying turbo-cartocss)?

fdansv:
> Should we add this also to torque layers?
I don't know, do we need it in Torque? They fill a very specific need for d3.cartodb

> One extra question, is this returning the processed cartocss (after applying turbo-cartocss)?
I don't think so, since the intent according to https://github.com/CartoDB/d3.cartodb/issues/117 is to preprocess turbocartocss on the client.

rochoa:
I guess it would be really helpful to have the preprocessed turbo-cartocss until we implement the client side datasource.

Another thing to consider is ... you can implement the datasource client side but that will work when you want to have a datasource based on visible data but it won't work if you want to use the whole dataset.

fdansv:
> Another thing to consider is ... you can implement the datasource client side but that will work when you want to have a datasource based on visible data but it won't work if you want to use the whole dataset.

That is assuming we don't use the SQL api :wink: But you're right. How would we pass the preprocessed CSS with the metadata? As you can see right now we're using `layer.options.cartocss`

rochoa:
You won't be able to use the SQL API if you are using a named map/private
tables.

Your questions is what I though we were solving :P.

On Mon, Mar 7, 2016 at 11:24 AM, Francisco Dans <notifications@github.com>
wrote:

> Another thing to consider is ... you can implement the datasource client
> side but that will work when you want to have a datasource based on visible
> data but it won't work if you want to use the whole dataset.
>
> That is assuming we don't use the SQL api [image: :wink:] But you're
> right. How would we pass the preprocessed CSS with the metadata? As you can
> see right now we're using layer.options.cartocss
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/CartoDB/Windshaft/pull/457#issuecomment-193195427>.
>



-- 
Raul Ochoa
rochoa@cartodb.com

fdansv:
The goal of this PR is just passing the cartocss style so that named maps can be rendered on the client. I'm not really dealing with turbocartocss here.

Related: https://github.com/CartoDB/cartodb.js/pull/1177#discussion_r55186856

rochoa:
But we have to deal with it.

On Mon, Mar 7, 2016 at 12:27 PM, Francisco Dans <notifications@github.com>
wrote:

> The goal of this PR is just passing the cartocss style so that named maps
> can be rendered on the client. I'm not really dealing with turbocartocss
> here.
>
> Related: CartoDB/cartodb.js#1177 (comment)
> <https://github.com/CartoDB/cartodb.js/pull/1177#discussion_r55186856>
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/CartoDB/Windshaft/pull/457#issuecomment-193213463>.
>



-- 
Raul Ochoa
rochoa@cartodb.com

fdansv:
As discussed offline, turbocartocss will be passed from Windshaft-CartoDB as explained in https://github.com/CartoDB/Windshaft/issues/453. Merging this to fix named maps.

PR #458, opened 2016-03-07
Geojson needed properties

### LINKED ISSUE #450: Geojson-renderer should send only the needed columns
Some queries may return rows with a high number of columns, for geojson-renderer, this causes performance issues sending a big amount of data for each tile. We should filter and send only those columns that are required, e.g:
- columns used in CartoCSS: `[ irrelevant_column < 10]{ marker-fill: #fabada }`
- interactivity columns
- columns required to use with widgets

dgaubert:
Fix #450

cc @rochoa


lib/windshaft/utils/cartocss_utils.js
```diff
@@ -0,0 +1,46 @@
+'use strict';
+
+module.exports.getColumnNamesFromCartoCSS = function (cartocss) {
+    // filter out repeated columnNames
+    return getColumnNames(cartocss).filter(function (item, pos, self) {
+        return self.indexOf(item) === pos;
+    });
+};
+
+function getColumnNames (cartocss) {
```
rochoa: Is it not possible to use `carto` module for this? I would like to avoid using custom regexp for this.

dgaubert: Where can I fetch this module?

package.json
```diff
@@ -1,59 +1,62 @@
 {
```
rochoa: :sob:

dgaubert: Done, npm reformats tabs when I perform `npm install <whatever> --save`. :sob:

lib/windshaft/utils/cartocss_utils.js
rochoa: Actually I would like to have a function where I pass a layer and I know if there are missing columns. If you refine that I would split it into a function that given a layer from https://github.com/CartoDB/Windshaft/blob/master/doc/MapConfig-1.5.0.md it returns the required columns from: geom_column, cartocss, interactivity, and attributes and another that validates the query associated is returning all those required functions.

lib/windshaft/utils/cartocss_utils.js
```diff
@@ -0,0 +1,66 @@
+'use strict';
+var carto = require('carto');
+var torque_reference = require('torque.js').cartocss_reference;
+
+module.exports.getColumnNamesFromCartoCSS = function (cartocss) {
+    return selectors(cartocss);
+};
+
+carto.tree.Reference.setData(torque_reference.version.latest);
```
javisantana: It would be pretty nice to have these in carto repo, not asking to change, just to let you know

lib/windshaft/utils/cartocss_utils.js
```diff
@@ -0,0 +1,66 @@
+'use strict';
+var carto = require('carto');
+var torque_reference = require('torque.js').cartocss_reference;
+
+module.exports.getColumnNamesFromCartoCSS = function (cartocss) {
```
javisantana: in the editor we do in this way https://github.com/CartoDB/cartodb/blob/master/lib/assets/javascripts/cartodb/models/carto.js#L1000

lib/windshaft/utils/geojson_utils.js
```diff
@@ -0,0 +1,59 @@
...
+
+module.exports.getGeojsonProperties = function (layerOptions) {
+    var properties = cartocssUtils.getColumnNamesFromCartoCSS(layerOptions.cartocss);
+
+    properties = properties.concat(getColumnNamesFromInteractivity(layerOptions.interactivity));
+    properties = properties.concat(getColumnNamesFromWidgets(layerOptions.widgets));
+    properties = properties.concat(getColumnNamesFromFilters(layerOptions.filters));
+
+    // filter out repeated ones
+    properties = properties.filter(function (item, pos, self) {
```
javisantana: not sure if you don't want to use underscore but you could use `_.uniq`

PR #462, opened 2016-03-15
Adds cartocss to metadata of layergroup

### LINKED ISSUE #452: CartoCSS should be sent in metadata

dgaubert:
Fixes #452

cc/ @rochoa @fdansv


lib/windshaft/metadata/layer-metadata.js
```diff
@@ -47,6 +47,12 @@ LayerMetadata.prototype.getMetadata = function (rendererCache, params, mapConfig
             if (_.size(widgets)) {
                 metadata[layerId].widgets = widgets;
             }
+
+            var cartocss = mapConfig.getCartoCSS(layerId);
+
+            if (_.size(cartocss)) {
```
rochoa: Why do we use `_` for this?

lib/windshaft/models/mapconfig.js
```diff
@@ -167,6 +171,10 @@ MapConfig.prototype.getLayers = function() {
 };
 
 
+MapConfig.prototype.getCartoCSS = function(layerIndex) {
```
rochoa: Let's ship it like this, but I don't like the fact that we are exposing a function that only makes sense for some type of layers.

dgaubert: Makes sense, but I don't see any problem in that. I mean, it works like an abstract method, in some cases it applies and some other cases it doesn't. Probably I'm missing something.

lib/windshaft/metadata/layer-metadata.js
dgaubert: I've just copied `widgets` way. But yes, I don't like it.

PR #463, opened 2016-03-15
Use carto parser to extract relevant selectors from cartocss

rochoa:
cc @dgaubert


dgaubert:
:+1: :rocket:

Thanks!

PR #465, opened 2016-03-23
[dnm] Simplify geometries keeping at least their bbox

rochoa:
Multigeometries should retain all their geometries but simplified.


rochoa:
cc @fdansv @javisantana @dgaubert

lib/windshaft/renderers/mapnik/geojson_template.sql
```diff
@@ -17,10 +27,10 @@ __simplified_geometries AS (
         ){{?}},
         {{= it.xyzResolution }} * {{= it.simplifyDpRatio}}
     ) __the_geometry
-    FROM ({{= it.layerSql }}) AS __cdb_query
+    FROM __dumped
```
javisantana: would it be possible to do the intersection in the previous step, I think it will be better since there are less BBOX to check (internally postgis saves the bbox for the whole geometry so when you dump I don't know if the bbox for each geometries are recreated)

rochoa: Oh, totally. I missed that one! I will speed up a lot because we will skip a lot of st_dumps.

rochoa:
All tests are fixed now without modifying anything in the query.

lib/windshaft/renderers/mapnik/geojson_template.sql
```diff
@@ -17,26 +36,36 @@ __simplified_geometries AS (
...
 ),
-__filtered_geometries AS (
+__collected_geometries AS (
     SELECT {{ if (it.columns && it.columns.length > 0) { }}
         {{= it.columns }},
     {{ } }}
-    CASE WHEN ST_IsEmpty(__the_geometry) OR __the_geometry IS NULL
-        THEN null::geometry
-        ELSE __the_geometry
+    CASE WHEN ST_NPoints({{= it.geomColumn }}) = 1
```
rochoa: The problem with this approach is that `LineString`s are converted into `MultiLineString`s, I gonna try ST_NumGeometries as an alternative.

PR #466, opened 2016-03-23
Do not direct access to array elements, iterate over them

rochoa:
cc @fdansv @dgaubert

### DIFF
--- a/lib/windshaft/utils/cartocss_utils.js
+++ b/lib/windshaft/utils/cartocss_utils.js
@@ -51,11 +51,14 @@ function appendRulesFields(definitions, allSelectors) {
         .reduce(function(allRules, rules) {
             return allRules.concat(rules);
         }, [])
-        .filter(function(rule) {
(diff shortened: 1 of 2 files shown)


lib/windshaft/utils/cartocss_utils.js
```diff
@@ -51,11 +51,14 @@ function appendRulesFields(definitions, allSelectors) {
         .reduce(function(allRules, rules) {
```
dgaubert: It's equivalent a `map`, isn't it?

dgaubert: I mean, something like that:
```
.map(function (rules) {
  return rules;
})
```

dgaubert: Ok. I see, we need to build a flat array.

rochoa: Yep, that's it. We need an array with all rules. With a map we would have an array with arrays.

PR #473, opened 2016-04-27
Geojson dataview columns

dgaubert:
Fixes [Windshaft-cartodb#427](https://github.com/CartoDB/Windshaft-cartodb/issues/427)

cc @rochoa


rochoa:
I understand the complexity of abstracting this so windshaft-cartodb could add what columns should be used so for now I consider this fair enough. :shipit:

PR #474, opened 2016-04-27
Handled "columns" property in layer options of map-config

dgaubert:
cc @rochoa


rochoa:
I don't know if it's worthy to check `options.columns` is an array.

rochoa:
Apart from that :shipit:.

PR #475, opened 2016-04-28
Adds support for buffer-size CartoCSS Map property in geojson tiles

### LINKED ISSUE #422: Add support for buffer-size in geojson renderer
geojson tiles should support cartocss' buffer-size.

rochoa:
Closes #422

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,7 +1,11 @@
 # Version 1.18.1
 2016-mm-dd
 
+New features:
  - Adds support for buffer-size CartoCSS Map property in geojson tiles.
+
+Announcements:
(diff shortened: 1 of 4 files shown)


rochoa:
cc @dgaubert

dgaubert:
Apart from the usage of `eval`, I guess that is totally necessary, LGTM

rochoa:
The method name is a bit unfortunate, but it's not the global `eval` thingy: https://github.com/CartoDB/carto/blob/176886f1ad4b66f74bd95fa24e59cb60df1962c4/lib/carto/renderer_js.js#L39-L43

dgaubert:
OMG 😱 
Also `jshint` scares

👍 then!

PR #491, opened 2016-08-04
show & hide mapnik layers

dgaubert:
Implements [Windshaft-cartodb#550](https://github.com/CartoDB/Windshaft-cartodb/issues/550). Allows to show/hide layers by requesting the different layer indexes in the URL.

*Note*: It's still pending to use layer id as an alternative to current layer index


dgaubert:
Mapnik layerId filtering is done. 

I think we can CR.

dgaubert:
I fixed layer interactivity issue. I'm ready for CR. 

cc @rochoa

lib/windshaft/models/mapconfig.js
```diff
@@ -127,6 +127,26 @@ function getLayerId(rawMapConfig, layerIndex) {
     return layerId;
 }
 
+MapConfig.prototype.getLayerById = function (layerId) {
```
rochoa: I don't find any usage of this one.

lib/windshaft/models/mapconfig.js
```diff
@@ -127,6 +127,26 @@ function getLayerId(rawMapConfig, layerIndex) {
...
+    var layers = this.getLayers();
+
+    for (var i = 0; i < layers.length; i++) {
+        if (layers[i].id === layerId) {
+            return layers[i];
+        }
+    }
+};
+
+MapConfig.prototype.getIndexByLayerId = function (layerId) {
```
rochoa: Should we return `-1` when the layer is not found? As when you use `Array.prototype.indexOf`.

PR #499, opened 2016-11-05
Fix map validator for basemaps

dgaubert:


rochoa:
This is an incomplete solution. Just consider the following example:

```
{
    "layers": [
        {
            "id": "globals",
            "options": {
                "cartocss": "@landmass_fill: #0e0e0e;",
                "cartocss_version": "2.3.0",
                "sql": "SELECT null::geometry AS the_geom_webmercator LIMIT 0"
            },
            "type": "mapnik"
        },
        {
            "id": "borders",
            "options": {
                "cartocss": "#background { polygon-fill: @landmass_fill; }",
                "cartocss_version": "2.3.0",
                "sql": "SELECT the_geom_webmercator FROM world_borders"
            },
            "type": "mapnik"
        },
        {
            "id": "capitals",
            "options": {
                "cartocss": "#labels { text-face-name: \"INVALID FONT\"; }",
                "cartocss_version": "2.1.1",
                "sql": "SELECT * FROM world_capitals"
            },
            "type": "mapnik"
        }
    ]
}
```

The `"INVALID FONT"` in `"id": "capitals"` will raise a general error, but later when validating layers individually the error presented to the client will be related to `"id": "borders"` as it won't be able to resolve the `@landmass_fill` global variable.

dgaubert:
How could we resolve `@landmass_fill` global variable to validate one layer??

Issue #503, opened 2016-11-09
Does windshaft now depend on cartodb enabled databases ?


strk:
I get this error upon sending a map config to Windshaft:
```
{"errors":["function cdb_querytablestext(unknown) does not exist"]}
```

Does this mean Windshaft now *requires* a cartodb-enabled database ?
( this wasn't the case until "recently" )

\cc @rochoa @pramsey

strk:
The dep is here:
```
lib/windshaft/utils/query_utils.js:20:        'SELECT * FROM CDB_QueryTablesText($windshaft$',
```

I'm looking at `Windshaft` rather than `Windshaft-CartoDB` to do w/out the whole `CartoDB` infrastructure, is this still a possibility ?

javisantana:
you need to install cartodb-postgres, that's the only dependency

strk:
On Wed, Nov 09, 2016 at 10:31:08AM -0800, javi santana wrote:
> you need to install cartodb-postgres, that's the only dependency

Then it would be worth adding it to the 

# Dependencies

section in the README file

( so far I found out it's just the function in 2 files included
  in the Windshaft testsuite, but I might have missed something )

javisantana:
I think we have those files just to pass tests.

we accept PR's as you probably know

rochoa:
What version are you using?

What do you mean by:
> I get this error upon sending a map config to Windshaft

That function is used optionally based on configuration.

strk:
@rochoa version 2.6.2. By `sending a map config` I mean hitting the `:base_url_mapconfig` endpoint with a `MapConfig` json

strk:
PR ready: #505

rochoa:
Can you be more explicit about:

> hitting the :base_url_mapconfig endpoint with a MapConfig json

BTW, have in mind Windshaft is no longer a service/server, it's just a library as stated in the documentation.

strk:
@rochoa the code using the `cartodb-postgresql` function is in the library:
```
lib/windshaft/utils/query_utils.js:20:        'SELECT * FROM CDB_QueryTablesText($windshaft$',
```

I hadn't researched the exact code path hitting that point, but I'm using the `example/server.js` implementation

strk:
The function in query_utils.js:20 (extractTableNames) is called by MapnikLayerMetadata.prototype.getMetadata in
lib/windshaft/metadata/mapnik-layer-metadata.js:36

[...]

PR #505, opened 2016-11-10
Split dependencies section between server dependency and datasource ones

strk:
See #503

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,5 @@
-Windshaft map tiler
+
+indshaft map tiler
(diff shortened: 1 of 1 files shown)


strk:
Would be even better to drop those dependencies, but until it's dropped, it's better to have it documented...

rochoa:
The `CDB_QueryTables` dependency, and thus `cartodb-postgresql`, is an optional one. So I see no need to be that picky in the README.

PR #506, opened 2016-11-10
Use "affected_table" in mapnik layer option when available

strk:
.. instead of extracting table names from the query (which is known
to not always yeld the correct result, especially when the query
uses functions to fetch data)

Also helps reducing dependency on cartodb-postgresql for deploys
with controlled MapConfig editing (see #503)

NOTE: this was tested on a live system, but the PR does not contain an automated test


rochoa:
Thanks a lot @strk!

Issue #479, opened 2016-05-28
Why have you pegged the mapnik dependency to some random Github commit?


tjwebb:
https://github.com/CartoDB/Windshaft/blob/master/package.json#L30

Is widshaft compatible with newer node-mapnik versions such as 3.5.x? In general, is the logic of windshaft so closely coupled with node-mapnik versions that it can't be a `peerDependency`?

tjwebb:
Same deal here: https://github.com/CartoDB/Windshaft/blob/master/package.json#L24.

This old version of canvas will not compile in any recent node version (>= 4.0). Is there something so specific about the CartoDB "1.2.7-cdb1" fork of canvas that makes it necessary to pin versions like this? It's crazy that I need to install an ancient version of node to make everything compile.

strk:
And same with tilelive-bridge and tilelive-mapnik. It would be very useful to know why custom version should be needed. When it was my turn to maintain this thing I was very frustrated by the impossibility to put comments inside the package.json. Maybe an idea could be to add a package.json.README file to contain notes about weird things found in there \cc @rochoa

Algunenano:
This is no longer true.

PR #531, opened 2017-02-09
Remove non-upstream dependencies

strk:
See #479


rochoa:
This will be superseded by #533, but it will still contain some forked dependencies.

PR #549, opened 2017-04-24
[tilelive-* release required] Use different buffer-size values for different tile formats

### LINKED ISSUE #548: Use different buffer-size values for different tile formats
We're going to need a way to handle different `buffer-size` values for different formats. For example, [this issue](https://github.com/CartoDB/Windshaft/issues/522) requests to have a default `buffer-size` of 0 for MVT format.

####  Tasks
- [ ] Add `buffer-size` property with different values for different formats in the Mapconfig. For example:
   ```
   buffersize: {
       png: 128,
       grid: 64,
       mvt: 0
   }
   ```
- [ ] Update Mapconfig documentation and release new version with the added changes

dgaubert:
Fixes #548

### DIFF
--- a/doc/MapConfig-1.6.0.md
+++ b/doc/MapConfig-1.6.0.md
@@ -1,0 +1,331 @@
+# 1. Purpose
+
+This specification describes [MapConfig](MapConfig-specification.md) format version 1.6.0.
+
+
+# 2. File format
+
+Layergroup files use the JSON format as described in [RFC 4627](http://www.ietf.org/rfc/rfc4627.txt).
+
+```javascript
+{
+    // OPTIONAL
+    // Version of this spec to use for validation.
+    // Defaults to "1.0.0".
+    version: "1.6.0",
+
+    // OPTIONAL
+    // default map extent, in map projection
+    // (only webmercator supported at this version)
+    extent: [-20037508.5, -20037508.5, 20037508.5, 20037508.5],
+
+    // OPTIONAL
+    // Spatial reference identifier for the map
+    // Defaults to 3857
+    srid: 3857,
+
+    // OPTIONAL
+    // maxzoom to be renderer. From this zoom tiles will respond 404
+    // default: undefined (infinite)
+    maxzoom: 18,
+
+    // OPTIONAL
+    // minzoom to be renderer. From this zoom tiles will respond 404. Must be less than maxzoom
+    // default: 0
+    minzoom: 3,
+
+    // OPTIONAL
+    // Extra tolerance around the map (in pixels) used to ensure labels crossing tile boundaries are equally rendered 
+    // in each tile (e.g. cut in each tile).
+    // 
+    // To configure `buffer-size` for any tile format: 
+    //  - buffersize: 128
+    //
+    // To configure `buffer-size` per tile format:
+    //  - buffersize: {
+    //      'png': 64,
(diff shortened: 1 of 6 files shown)


doc/MapConfig-1.6.0.md
```diff
@@ -0,0 +1,331 @@
...
+    // default: undefined (infinite)
+    maxzoom: 18,
+
+    // OPTIONAL
+    // minzoom to be renderer. From this zoom tiles will respond 404. Must be less than maxzoom
+    // default: 0
+    minzoom: 3,
+
+    // OPTIONAL
+    // Extra tolerance around the map (in pixels) used to ensure labels crossing tile boundaries are equally rendered
```
rochoa: I'm missing the documentation related to what priority this has. Does this supersede the buffer-size in the CartoCSS or the other way around?

doc/MapConfig-1.6.0.md
```diff
@@ -0,0 +1,331 @@
...
+    // minzoom to be renderer. From this zoom tiles will respond 404. Must be less than maxzoom
+    // default: 0
+    minzoom: 3,
+
+    // OPTIONAL
+    // Extra tolerance around the map (in pixels) used to ensure labels crossing tile boundaries are equally rendered 
+    // in each tile (e.g. cut in each tile).
+    // 
+    // To configure `buffer-size` for any tile format: 
+    //  - buffersize: 128
```
rochoa: What about simplifying this? Not a big fan of having different ways/types to set a configuration.

ethervoid: We're doing the same way `metatile` property is doing but we could just let the buffer-size definition by format. What do you think?

rochoa: I see 😄 . But with the metatile we already had the one with just one metatile for all formats and we added the _by format_ one.

So let's go with your proposal, just expose the _by format_ one, ok?

doc/MapConfig-1.6.0.md
```diff
@@ -0,0 +1,332 @@
...
+    // To configure `buffer-size` per tile format:
+    //  - buffersize: {
+    //      'png': 64,
+    //      'grid.json': 64,
+    //      'mvt': 0
+    //    }
+    //
+    // Buffer-size also can be configured through `cartocss`, in these cases the priority to apply the right `buffer-size` is:
+    //  - Raster: cartocss >> map-config >> default
+    //  - Vector: map-config >> default
```
rochoa: What happens if I define `Map { buffer-size: 64 }` in one of my layers' CartoCSS and `buffersize: {mvt: 0}` in here, and I retrieve a MVT tile? Do we ignore the buffer size defined through the CartoCSS property in that scenario?

ethervoid: Yes, we don't take into account the `buffer-size` value defined in CSS for the `mvt` tiles

dgaubert: Please note that it's the default behaviour in `tilelive-bridge`. For vector formats they create a map from the `xml` and then always override the `buffer-size` value, before our changes always to 256 and now with a custom value. On the contrary, `tilelive-mapnik` has another approach creating the map instance with a default value of `buffer-size` and then it's overridden. As we want to sent a PR to upstream for `tilelive-bridge` we didn't take into account to change that behaviour.

PR #554, opened 2017-07-25
[WIP] pg_mvt renderer for MVT

rafatower:
This is my implementation in Windshaft of a renderer that makes use of PostGIS `ST_AsMVT` function to render torque tiles directly in the database. As you can see, it is full of TODO's.

Based on torque renderer.

IMHO the current biggest limitations are:
- the additional feature attributes (additional columns) to be used when enconding are hardcoded.
- it only supports one layer. I don't know whether it makes sense to support more.
- there's no way to switch from mapnik renderer to `pg_mvt` renderer. This could be done through config I guess.

cc/ @Algunenano @javisantana


rochoa:
I guess, after merging https://github.com/CartoDB/Windshaft/pull/561, we no longer need this PR, right?

rochoa:
cc @dv343

dv343:
I think this can be safely deleted.

PR #561, opened 2017-09-28
PostgreSQL MVT path

dv343:
Adds a new path to generate MVT tiles without using Mapnik.

The new path needs PostGIS 2.4 since PostGIS is the one who generates the MVT tiles with the new path, with the St_AsMVT and St_AsMVTGeom functions.

Since complying with this requisite could be problematic, and we may want to rollback to the old functionality, a flag is used to switch between the old mapnik path and this new one.

The flag is `opts.mvt.usePgSQL`, and it is passed to the constructor by the upper layers. If it doesn't exists it will default to the old path.

Do not merge yet, since the tests are passing because they don't target the new functionality.


test/acceptance/torque.js
```diff
@@ -293,10 +293,10 @@ describe('torque', function() {
         var testClient = new TestClient(mapConfig);
         testClient.getTile(0, 0, 0, {layer: 0, format: 'torque.json'}, function(err, torqueTile) {
             SubstitutionTokens.replace = replaceFn;
-
             assert.ok(!err, err);
-            assert.deepEqual(torqueTile, [{ x__uint8: 128, y__uint8: 128, vals__uint8: [2,3], dates__uint16: [1,0] }]);
-
+            torqueTile[0].vals__uint8.sort();
+            torqueTile[0].dates__uint16.sort();
+            assert.deepEqual(torqueTile, [{ x__uint8: 128, y__uint8: 128, vals__uint8: [2,3], dates__uint16: [0,1] }]);
```
rochoa: Why are these changes here?

lib/windshaft/backends/map_validator.js
```diff
@@ -100,9 +100,9 @@ MapValidatorBackend.prototype.validate = function(mapConfigProvider, callback) {
                     var layerType = mapConfig.layerType(layerId);
 
                     if (layerType === 'mapnik') {
-                        if ( lyropt.interactivity ) {
-                            validateRemainingFormatLayerFns.push(validateMapnikGridJson(layerId));
-                        }
+                         if ( lyropt.interactivity ) {
```
rochoa: Can we avoid code style changes to keep the PR as clear as possible?

lib/windshaft/renderers/mapnik/geojson_renderer.js
```diff
@@ -83,7 +83,11 @@ GeojsonRenderer.prototype._getTileData = function(cartoPSQL, coord, zoom, layer,
         // See https://github.com/mapnik/mapnik/wiki/ScaleAndPpi#scale-denominator
         scale_denominator: xyzResolution / 0.00028,
         pixel_width: xyzResolution,
-        pixel_height: xyzResolution
+        pixel_height: xyzResolution,
```
rochoa: Why is this here?

lib/windshaft/renderers/pg_mvt/factory.js
```diff
@@ -0,0 +1,123 @@
...
+///            http://github.com/CartoDB/CartoDB-SQL-API/blob/1.8.2/doc/API.md#json
+///            The 'readonly' parameter (false by default) requests
+///            that running the query should not allowed to change the database.
+///     - dbPoolParams: database connection pool params
+///          - size: maximum number of resources to create at any given time
+///          - idleTimeout: max milliseconds a resource can go unused before it should be destroyed
+///          - reapInterval: frequency to check for idle resources
+///
+function PgMvtFactory(options) {
+    // jshint newcap:false
```
rochoa: Do we need this jshint?

lib/windshaft/renderers/pg_mvt/factory.js
```diff
@@ -0,0 +1,123 @@
+var _ = require('underscore');
```
rochoa: Nowadays, it's more common to use `filename-with-dash-as-separator.js` format, could we change it here? 🙏

lib/windshaft/renderers/pg_mvt/factory.js
```diff
@@ -0,0 +1,123 @@
...
+///          - size: maximum number of resources to create at any given time
+///          - idleTimeout: max milliseconds a resource can go unused before it should be destroyed
+///          - reapInterval: frequency to check for idle resources
+///
+function PgMvtFactory(options) {
+    // jshint newcap:false
+    this.options = options || {};
+    _.defaults(this.options, {
+        // TODO we most likely do not need such adaptor
+        sqlClass: PSQLAdaptor(PSQL, options.dbPoolParams)
```
rochoa: Do we really need this `PSQLAdaptor` indirection? Couldn't we just use `PSQL` directly?

lib/windshaft/renderers/pg_mvt/factory.js
```diff
@@ -0,0 +1,123 @@
...
+
+    getAdaptor: function (renderer, format, onTileErrorStrategy) {
+        return new BaseAdaptor(renderer, format, onTileErrorStrategy);
+    },
+
+    getRenderer: function (mapConfig, format, options, callback) {
+        var dbParams = RendererParams.dbParamsFromReqParams(options.params);
+        var layer = options.layer;
+
+        if (!formatToRenderer[format]) {
```
rochoa: Wouldn't be better to call `this.supportsFormat`?

lib/windshaft/renderers/pg_mvt/factory.js
```diff
@@ -0,0 +1,123 @@
...
+
+        var mapLayers = mapConfig.getLayers();
+        mapLayers.forEach((layer, layerIndex) => {
+            layer.id = mapConfig.getLayerId(layerIndex);
+        });
+
+
+        var mapnik_layers = [];
+        var layerFilter = options.layer;
+        if (typeof layerFilter === 'number') {
```
rochoa: Why don't we rely on/implement this on `utils/layer_filter`?

lib/windshaft/renderers/pg_mvt/renderer.js
```diff
@@ -0,0 +1,77 @@
...
+                // See https://github.com/mapnik/mapnik/wiki/ScaleAndPpi#scale-denominator
+                scale_denominator: `(cdb_XYZ_Resolution(${z}) / 0.00028)`,
+                pixel_width: `cdb_XYZ_Resolution(${z})`,
+                pixel_height: `cdb_XYZ_Resolution(${z})`,
+                var_zoom: z,
+                var_x: x,
+                var_y: y
+            });
+
+            query += `coalesce((` +
```
rochoa: As we are already using template strings, I would avoid `+` concatenation when possible.

lib/windshaft/renderers/pg_mvt/renderer.js
```diff
@@ -0,0 +1,77 @@
...
+    /// @param z tile zoom
+    /// callback: will be called when done using nodejs protocol (err, data)
+    getTile: function (z, x, y, callback) {
+        var query = 'SELECT (';
+        this.layers.forEach((layer, i) => {
+            if (i >= 1) {
+                query += ' || ';
+            }
+            var nonGeoColumns = geojsonUtils.getGeojsonProperties(layer.options).join(',');
+            if (nonGeoColumns !== '') {
```
rochoa: I would put all columns, even the geometry one, in the array, so later you can just simply do `.join(',')`.

lib/windshaft/renderers/pg_mvt/renderer.js
```diff
@@ -0,0 +1,77 @@
...
+        this.sql(query, function (err, data) {
+            timer.end('query');
+            if (err) {
+                debug("Error running pg_mvt query " + query + ": " + err);
+                if (err.message) {
+                    err.message = "PgMvtRenderer: " + err.message;
+                }
+                callback(err);
+            } else {
+                callback(null, data.rows[0].st_asmvt, { 'Content-Type': 'application/x-protobuf' }, timer.getTimes());
```
rochoa: Should we sanity check `rows[0]. st_asmvt `?

lib/windshaft/renderers/renderer_factory.js
```diff
@@ -29,7 +32,10 @@ function RendererFactory(opts) {
             opts.http.fallbackImage
         )
     ];
-    this.factories = availableFactories.reduce(function(factories, factory) {
+    if (opts.mvt.usePgSQL) {
```
rochoa: Why not having it available all the time?

lib/windshaft/renderers/renderer_factory.js
```diff
@@ -29,7 +32,10 @@ function RendererFactory(opts) {
...
+        availableFactories.push(new PgMvtRenderer.factory(opts.torque)); // TODO abusing torque
```
rochoa: We cannot rely on torque options outside of torque.

test/acceptance/mvt.js
```diff
@@ -4,13 +4,13 @@ var assert = require('../support/assert');
 var TestClient = require('../support/test_client');
 var mapnik = require('mapnik');
 
-describe('mvt', function() {
```
rochoa: We are still testing with Mapnik implementation, aren't we?

test/acceptance/mvt.js
```diff
@@ -132,35 +137,35 @@ describe('mvt', function() {
...
+            testClient.getTile(13, 4011, 3088, { layer: 'mapnik', format: 'mvt' }, multipleLayersValidation(done));
         });
 
-        it('invalid mvt layer', function(done) {
+        it('invalid mvt layer', function (done) {
             var testClient = new TestClient(mixedLayersMapConfig);
-            testClient.getTile(13, 4011, 3088, { layer: 0, format: 'mvt'}, function(err) {
+            testClient.getTile(13, 4011, 3088, { layer: 0, format: 'mvt' }, function (err) {
                 assert.ok(err);
-                assert.equal(err.message, 'Unsupported format mvt');
```
rochoa: Why did we change the error message?

test/acceptance/mvt.js
```diff
@@ -4,13 +4,13 @@ var assert = require('../support/assert');
 var TestClient = require('../support/test_client');
```
rochoa: This file has too many changes related to code style. Could we revert all changes related to code style and fix them in another PR? It's hard to follow what it's actually changing.

test/acceptance/mvt.js
```diff
@@ -4,13 +4,13 @@ var assert = require('../support/assert');
...
 var mapnik = require('mapnik');
 
-describe('mvt', function() {
+describe('mvt', function () {
```
rochoa: I guess we need to configure this suite to run with both valued, true and false, for the usePgSQL configuration option, right?

test/acceptance/torque.js
dv343: I will push these changes in a separated PR.

lib/windshaft/renderers/mapnik/geojson_renderer.js
dv343: I will remove those changes, but the geojson is going to be removed anyway in another PR.

lib/windshaft/renderers/pg_mvt/factory.js
dv343: What do you mean?

lib/windshaft/renderers/pg_mvt/factory.js
dv343: No, fixed

lib/windshaft/renderers/pg_mvt/factory.js
dv343: Maybe, that code is yours and Rafa's I haven't touched it. You know more than me.

lib/windshaft/renderers/pg_mvt/factory.js
dv343: Changed

lib/windshaft/renderers/pg_mvt/factory.js
dv343: I relied on it partially, just like mapnik renderer does. It's true that that code asks for a refactor, but IMO it's outside this PR goals, and it's not particularly related to MVT. See https://github.com/CartoDB/Windshaft/blob/master/lib/windshaft/renderers/mapnik/factory.js#L157

lib/windshaft/renderers/pg_mvt/renderer.js
dv343: No, look at the query (https://github.com/CartoDB/Windshaft/pull/561/files/ff5ae340771e9fb180667128ad564d6614a3ff6b#diff-44ec6241c20bd52513803aa408451c69R56) Mixing columns would be a problem.

lib/windshaft/renderers/pg_mvt/renderer.js
dv343: If you do that all the newlines and spaces will be sent to the DB, IMO, it's better to avoid that.

lib/windshaft/renderers/pg_mvt/renderer.js
dv343: I don't think so. The query is already a workaround (for safety) for known PostGIS problems (which have been already addressed at PostGIS). If rows has zero length then the query is bugged or PostGIS behavior have changed, in any case the automatic exception will be Ok, IMO.

lib/windshaft/renderers/renderer_factory.js
dv343: For isolation, if the new path is disabled, then no code of the new path should be executed. If a bug is present in that part (at initialization time) then we don't need to deploy a fast fix, a configuration change would be enough.

Also, I think it's better to let pg_mvt renderer assume he has a PostGIS 2.4, if he hasn't then the factory shouldn't create him.

lib/windshaft/renderers/renderer_factory.js
dv343: I think this is related to the SQLAdapator, I think those options control DB connections. Again, I don't know about how we manage connections since I haven't programmed that part. It is also tightly coupled with windshaft-cartodb, so changing this requires a PR to windshaft-cartodb. I agree that using torque opts is ugly and should be fixed, but I need more information.

lib/windshaft/renderers/pg_mvt/factory.js
rochoa: `lib/windshaft/renderers/pg_mvt/factory.js` -> `lib/windshaft/renderers/pg-mvt/factory.js`.

lib/windshaft/renderers/pg_mvt/factory.js
rochoa: OK, fair enough.

lib/windshaft/renderers/pg_mvt/renderer.js
rochoa: Why?


```diff
diff --git a/lib/windshaft/renderers/pg_mvt/renderer.js b/lib/windshaft/renderers/pg_mvt/renderer.js
index 7726be26..6cf5ccdd 100644
--- a/lib/windshaft/renderers/pg_mvt/renderer.js
+++ b/lib/windshaft/renderers/pg_mvt/renderer.js
@@ -35,10 +35,8 @@ Renderer.prototype = {
             if (i >= 1) {
                 query += ' || ';
             }
-            var nonGeoColumns = geojsonUtils.getGeojsonProperties(layer.options).join(',');
-            if (nonGeoColumns !== '') {
-                nonGeoColumns += ',';
-            }
+            var columns = geojsonUtils.getGeojsonProperties(layer.options);
+            columns.push(`ST_AsMVTGeom(the_geom_webmercator, CDB_XYZ_Extent(${x},${y},${z}), ${this.mvt_extent}, 0, true)`);

             var subQuery = SubstitutionTokens.replace(layer.options.sql, {
                 bbox: `CDB_XYZ_Extent(${x},${y},${z})`,
@@ -53,7 +51,7 @@ Renderer.prototype = {

             query += `coalesce((` +
                 `select st_asmvt(geom, '${layer.id}') FROM` +
-                `(SELECT ${nonGeoColumns} ST_AsMVTGeom(the_geom_webmercator, CDB_XYZ_Extent(${x},${y},${z}), ${this.mvt_extent}, 0, true)` +
+                `(SELECT ${columns.join(',')}` +
                 `FROM (${subQuery}) AS cdbq where the_geom_webmercator && CDB_XYZ_Extent(${x},${y},${z}) ) as geom where geom is not null` +
                 `), '')`;
         });
```

test/acceptance/mvt.js
dv343: Changed

lib/windshaft/renderers/pg_mvt/renderer.js
rochoa: Is that a problem? The `subQuery` can contain anything, and it is also sent to the db.

lib/windshaft/backends/map_validator.js
dv343: Changed

test/acceptance/mvt.js
dv343: No, we are not, see last pushed version

lib/windshaft/renderers/pg_mvt/renderer.js
rochoa: What do you mean by _automatic exception_?

lib/windshaft/renderers/renderer_factory.js
rochoa: I don't understand this. Making the factory available doesn't mean it has to be used, that's why the `renderer_factory.js` has:

```js
if (format === 'mvt' && this.opts.mvt.usePgSQL) {
    return 'pg_mvt';
}
```

lib/windshaft/renderers/renderer_factory.js
rochoa: It's not about the connections, but about the params needed to create the connection. If we need to change Windshaft-cartodb, that's OK. But from a library standpoint of view, we cannot say: "_Hey, for MVT we need the Torque options_", that's not clear. So let's fix this.

test/acceptance/torque.js
dv343: Changed

lib/windshaft/backends/map_validator.js
dv343: Changed

lib/windshaft/renderers/mapnik/geojson_renderer.js
dv343: Changed

lib/windshaft/renderers/pg_mvt/factory.js
```diff
@@ -0,0 +1,121 @@
...
+///     - dbPoolParams: database connection pool params
+///          - size: maximum number of resources to create at any given time
+///          - idleTimeout: max milliseconds a resource can go unused before it should be destroyed
+///          - reapInterval: frequency to check for idle resources
+///
+function PgMvtFactory(options) {
+    this.options = options || {};
+    _.defaults(this.options, {
+        // TODO we most likely do not need such adaptor
+        sqlClass: new PSQLAdaptor(PSQL, options.dbPoolParams)
```
rochoa: My previous comment is missing. I would remove `PSQLAdaptor` and use `PSQL` directly. The `PSQLAdaptor` was part of torque renderer at the very beginning, but it doesn't make sense here anymore.

lib/windshaft/renderers/pg_mvt/factory.js
dv343: We don't need it, changed.

test/acceptance/mvt.js
```diff
@@ -133,33 +146,33 @@ describe('mvt', function() {
...
+            var testClient = new TestClient(mixedLayersMapConfig, options);
             testClient.getTile(13, 4011, 3088, { layer: 'mapnik', format: 'mvt'}, multipleLayersValidation(done));
         });
 
         it('invalid mvt layer', function(done) {
-            var testClient = new TestClient(mixedLayersMapConfig);
+            var testClient = new TestClient(mixedLayersMapConfig, options);
             testClient.getTile(13, 4011, 3088, { layer: 0, format: 'mvt'}, function(err) {
                 assert.ok(err);
-                assert.equal(err.message, 'Unsupported format mvt');
```
rochoa: My previous comment is missing: Why did we change the error message?

lib/windshaft/renderers/pg_mvt/renderer.js
dv343: You are mixing now raw column names with SQL expressions, but Ok.

lib/windshaft/renderers/pg_mvt/factory.js
dv343: Ok, I will change it.

lib/windshaft/renderers/pg_mvt/renderer.js
dv343: Changed

lib/windshaft/renderers/pg_mvt/renderer.js
dv343: I mean that JS will raise an exception at that `[0].st_asmvt` indirection, right?

lib/windshaft/renderers/renderer_factory.js
dv343: Invoking the constructor is using it, and having it available is invoking the constructor.

lib/windshaft/renderers/renderer_factory.js
dv343: It would be Ok to set all those params at `renderer.dbparams` (in terms of windshaft-cartodb)?

lib/windshaft/renderers/renderer_factory.js
rochoa: You are just making the factory available. It not a heavy class, and it's done just once.

lib/windshaft/renderers/pg_mvt/renderer.js
rochoa: It will, but you are not handling it. And you are in async context, meaning that the _client_ expecting to be called back won't get anything.

lib/windshaft/renderers/pg_mvt/factory.js
dv343: My comment is missing too. That code is yours and Rafa's, I don't know about it, what was its purpose? Why is it safe to remove it?

test/acceptance/mvt.js
dv343: I think that including the invalid format gives more information.

lib/windshaft/renderers/pg_mvt/factory.js
dv343: Changed

lib/windshaft/renderers/renderer_factory.js
dv343: Changed

lib/windshaft/renderers/pg-mvt/factory.js
```diff
@@ -0,0 +1,121 @@
...
+///     - dbPoolParams: database connection pool params
+///          - size: maximum number of resources to create at any given time
+///          - idleTimeout: max milliseconds a resource can go unused before it should be destroyed
+///          - reapInterval: frequency to check for idle resources
+///
+function PgMvtFactory(options) {
+    this.options = options || {};
+    _.defaults(this.options, {
+        // TODO we most likely do not need such adaptor
+        sqlClass: new PSQLAdaptor(PSQL, options.dbPoolParams)
```
rafatower: Instead of using the torque `PSQLAdaptor` I'd try to replace it by `PSQL` or see if we really need anything specific from it (I don't think so) and  replace it by its own adaptor, if needed.

lib/windshaft/renderers/renderer_factory.js
```diff
@@ -13,6 +14,8 @@ function RendererFactory(opts) {
     opts.http = opts.http || {};
     opts.mapnik = opts.mapnik || {};
     opts.torque = opts.torque || {};
+    opts.mvt = opts.mvt || {};
+    this.opts = opts;
```
rafatower: why is this line required?

lib/windshaft/renderers/renderer_factory.js
```diff
@@ -27,9 +30,10 @@ function RendererFactory(opts) {
             opts.http.timeout,
             opts.http.proxy,
             opts.http.fallbackImage
-        )
+        ),
+        new PgMvtRenderer.factory(opts.torque)
```
rafatower: I think this was already commented but yet: are we relying on torque options? I think it should stick to `opts.mvt`

lib/windshaft/renderers/renderer_factory.js
```diff
@@ -86,7 +90,11 @@ function genericMakeRenderer(factory, mapConfig, params, limits, onTileErrorStra
     );
 }
 
-function getFactoryName(mapConfig, layer, format) {
+RendererFactory.prototype.getFactoryName = function (mapConfig, layer, format) {
+    if (this.opts.mvt.usePostGIS && format === 'mvt' && isMapnikFactory(mapConfig, layer, format)) {
```
rafatower: why is the `isMapnikFactory` check needed?

lib/windshaft/renderers/pg-mvt/factory.js
```diff
@@ -0,0 +1,122 @@
...
+    supportsFormat: function (format) {
+        return !!formatToRenderer[format];
+    },
+
+    getAdaptor: function (renderer, format, onTileErrorStrategy) {
+        return new BaseAdaptor(renderer, format, onTileErrorStrategy);
+    },
+
+    getRenderer: function (mapConfig, format, options, callback) {
+        // jshint maxcomplexity: 7
```
rochoa: BTW, we should not introduce this in new code.

lib/windshaft/renderers/renderer_factory.js
dv343: Because `getFactoryName` needs to know if the PostGIS path is enabled

lib/windshaft/renderers/renderer_factory.js
dv343: Yes, we should! I didn't see this reference, good catch

lib/windshaft/renderers/renderer_factory.js
dv343: Well... this comes from a long discussion with @rochoa The way I see it we had 2 options: being consistent or doing it "right". They are mutually exclusive, and we decided to go with consistent. That function takes care of filtering.

rochoa: MVT format only make sense for layers where you have an SQL option (type:mapnik|cartodb). With this, we avoid trying to generate MVT tiles from `plain` or `http` layers, where it does not make sense at all.

lib/windshaft/renderers/pg-mvt/factory.js
dv343: I've just removed it

lib/windshaft/renderers/pg-mvt/factory.js
dv343: Fixed

.travis.yml
```diff
@@ -1,21 +1,13 @@
...
-  - NPROCS=1 JOBS=1 PGUSER=postgres CXX=g++-4.9
+services:
+  - docker
 
-language: node_js
-node_js:
-  - "6"
-dist: precise
+before_install:
+  - docker pull dmanzanares/windshaft-testing
```
rafatower: I don't feel 100% comfortable downloading from anyone's personal docker hubs (or any other repository). Don't know if there's any quick solution for this

Dockerfile
```diff
@@ -0,0 +1,31 @@
...
+# Configure PostgreSQL
+RUN echo "local     all       all                     trust" > /etc/postgresql/9.5/main/pg_hba.conf
+RUN echo "host      all       all       0.0.0.0/0     trust" >> /etc/postgresql/9.5/main/pg_hba.conf
+RUN echo "listen_addresses='*'" >> /etc/postgresql/9.5/main/postgresql.conf
+
+RUN apt-get -y purge $DEV_DEPENDENCIES
+RUN apt-get -y autoremove
+
+WORKDIR /srv
+CMD /etc/init.d/postgresql start && node -v && export NPROCS=1 && export JOBS=1 && export CXX=g++-4.9 && npm install && export PGUSER=postgres && npm test
```
rafatower: :clap: :clap: vamos!

lib/windshaft/renderers/renderer_factory.js
rafatower: understood, thanks for the clarification.

dv343:
From my point of view this branch is Ok now. However, it needs PostGIS 2.4 to use the new MVT path. If PostGIS 2.4 is not present the `opts` param should reflect that by disabling the new path.

The tests are ran in a docker xenial environment that compiles PostGIS 2.4 from sources.

lib/windshaft/renderers/pg-mvt/renderer.js
```diff
@@ -0,0 +1,83 @@
...
+            });
+
+            query += `(
+                select st_asmvt(geom, '${layer.id}') FROM
+                (SELECT ${nonGeoColumns}
+                    ST_AsMVTGeom(the_geom_webmercator, CDB_XYZ_Extent(${x},${y},${z}), ${this.mvt_extent}, 0, true)
+                FROM
+                    (${subQuery}) AS cdbq where the_geom_webmercator && CDB_XYZ_Extent(${x},${y},${z}) )
+                    AS geom
+                )`;
```
rafatower: pretty neat after removing the workarounds :)

.travis.yml
dv343: That's true, and it doesn't seem very professional... We should have a Docker organization with permissions and all that stuff... But at least the Dockerfile used to generate the image is in the repo.

lib/windshaft/renderers/pg-mvt/renderer.js
dv343: That's because the optimizations are missing :joy:

.travis.yml
javisantana: I've created a company account

javisantana:
what does this PR needs to be merged?

dv343:
@javisantana 
It needs a refactor to fix a bug in the filtering: https://github.com/CartoDB/Windshaft/pull/564 and partially merge https://github.com/CartoDB/Windshaft/tree/mvt-bug to check that we are downloading only those columns.

PR #564, opened 2017-10-04
Unify layer filtering logic in layer_filter

rochoa:
It handles aliases: all, mapnik, etc. It also casts numbers accordingly.

### DIFF
--- a/lib/windshaft/renderers/mapnik/factory.js
+++ b/lib/windshaft/renderers/mapnik/factory.js
@@ -270,7 +270,6 @@ MapnikFactory.prototype.getBufferSize = function(mapConfig, format, options) {
 };
 
 
-// TODO: fix this
 MapnikFactory.prototype.mapConfigToMMLBuilderConfig = function(mapConfig, queryRewriter, layerFilter) {
     var self = this;
     var options = {
--- a/lib/windshaft/renderers/mapnik/factory.js
+++ b/lib/windshaft/renderers/mapnik/factory.js
@@ -288,10 +288,7 @@ MapnikFactory.prototype.mapConfigToMMLBuilderConfig = function(mapConfig, queryR
     };
 
     var filteredLayerIndexes = layersFilter(mapConfig, layerFilter);
-    filteredLayerIndexes.filter(function (layerIndex) {
-        return mapConfig.layerType(layerIndex) === 'mapnik';
-    })
-    .reduce(function (options, layerIndex) {
+    filteredLayerIndexes.reduce(function (options, layerIndex) {
         var layer = mapConfig.getLayer(layerIndex);
 
         validateLayer(layer, layerIndex);
--- a/lib/windshaft/renderers/mapnik/factory.js
+++ b/lib/windshaft/renderers/mapnik/factory.js
@@ -288,7 +288,7 @@ MapnikFactory.prototype.mapConfigToMMLBuilderConfig = function(mapConfig, queryR
     };
 
     var filteredLayerIndexes = layersFilter(mapConfig, layerFilter);
-    filteredLayerIndexes.filter(layerIndex => {
+    filteredLayerIndexes.filter(function (layerIndex) {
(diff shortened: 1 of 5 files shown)


lib/windshaft/renderers/mapnik/factory.js
```diff
@@ -289,20 +287,8 @@ MapnikFactory.prototype.mapConfigToMMLBuilderConfig = function(mapConfig, queryR
...
-    } catch (err) {
-        // aliases mapnik, all,
-        filteredLayerIndexes = [];
-        for (var i = 0; i < mapConfig.getLayers().length; i++) {
-            filteredLayerIndexes.push(i);
-        }
-    }
-
-    filteredLayerIndexes.filter(function (layerIndex) {
-        return mapConfig.layerType(layerIndex) === 'mapnik';
```
dv343: This `filter` was wrong or it is expected to silently ignore non-supported layers?

rochoa: I'm relying on the fact that `layersFilter` is already being used in the RendererFactory, but good point, we might need to keep that filtering.

I'm also realising that `var filteredLayerIndexes = layersFilter(mapConfig, layerFilter);` might throw an exception, and I'm not handling it here.

rochoa: Although there is another `throw` within this function, so I guess we can keep the `layersFilter` call without a try/catch block.

lib/windshaft/renderers/renderer_factory.js
```diff
@@ -98,89 +98,25 @@ RendererFactory.prototype.getFactoryName = function (mapConfig, layer, format) {
...
     }
 
-    var factoryName = mapConfig.layerType(layer);
-
-    if (!factoryName) {
-        throw new Error('Invalid layer filtering');
-    }
-
-    return factoryName;
+    return BlendRenderer.factory.NAME;
```
dv343: There is an important difference here, with the new code, it won't throw an exception with an invalid configuration... Not here at least. Could you explain how is that managed?

lib/windshaft/renderers/renderer_factory.js
```diff
@@ -98,89 +98,25 @@ RendererFactory.prototype.getFactoryName = function (mapConfig, layer, format) {
...
-        return true;
-    }
-
-    if (layersFilter.isFilterByLayerId(layer) && isEveryMapnikByLayerId(mapConfig, layer)) {
-        return true;
-    }
-
-    return false;
+    var filteredLayers = layersFilter(mapConfig, layer);
+    return filteredLayers.every(isMapnikLayerFn(mapConfig));
```
dv343: Instead of using `isMapnikLayerFn` to return a function, can't we just map `filteredLayers` to get a list of the filtered layer types and then use the `every`?

lib/windshaft/renderers/renderer_factory.js
rochoa: The `layersFilter` will throw an error if filtering is not OK.

lib/windshaft/renderers/renderer_factory.js
rochoa: It's more or less the same, isn't it?

lib/windshaft/utils/layer_filter.js
```diff
@@ -1,85 +1,88 @@
...
+            filteredLayers.push(index);
+        }
+        return filteredLayers;
+    }, []);
+}
 
-    filteredLayers = filteredLayers.sort(function(a, b) { return a - b; });
 
+function resolveIds(mapConfig, layerIds) {
+    const layerIdsToNumber = layerIds.map(_ => +_);
```
dv343: I understand that we didn't support mixing layerIDs with layer indices, right?

rochoa: I thought we supported it, but from the tests, it's obvious we don't.

PR #568, opened 2017-10-06
Better docker

dv343:
Minor dependencies added to improve flexibility.

Removal of the `expose 5858` to avoid clashing between multiple dockers on host

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,9 +1,10 @@
 FROM ubuntu:xenial
 
-ENV DEV_DEPENDENCIES curl wget protobuf-c-compiler
-
+# Add external repos
 RUN apt-get -y update
-RUN apt-get -y install $DEV_DEPENDENCIES
+RUN apt-get install -y software-properties-common
+RUN add-apt-repository -y ppa:ubuntu-toolchain-r/test
+RUN apt-get -y update
 
 #Use UTF8 to avoid encoding problems with pgsql
 RUN apt-get -y install locales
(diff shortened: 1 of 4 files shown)


Dockerfile
```diff
@@ -12,19 +13,22 @@ RUN locale-gen en_US.UTF-8
...
 
 # Configure PostgreSQL
 RUN echo "local     all       all                     trust" > /etc/postgresql/9.5/main/pg_hba.conf
 RUN echo "host      all       all       0.0.0.0/0     trust" >> /etc/postgresql/9.5/main/pg_hba.conf
+RUN echo "host      all       all       ::1/128       trust" >> /etc/postgresql/9.5/main/pg_hba.conf
 RUN echo "listen_addresses='*'" >> /etc/postgresql/9.5/main/postgresql.conf
 
-RUN apt-get -y purge $DEV_DEPENDENCIES
+# Clean
+RUN apt-get -y remove wget protobuf-c-compiler
```
rochoa: Is this really necessary? If so, why don't we keep a list of dev dependencies as we had before? For instance, it's clear to me that we need to keep `curl` because some scripts use it to configure the database for the tests, but it's not clear in this context where you remove wget but keep curl.

lib/windshaft/renderers/pg-mvt/renderer.js
```diff
@@ -43,7 +43,7 @@ Renderer.prototype = {
             var subQuery = SubstitutionTokens.replace(layer.options.sql, {
                 bbox: `CDB_XYZ_Extent(${x},${y},${z})`,
                 // See https://github.com/mapnik/mapnik/wiki/ScaleAndPpi#scale-denominator
-                scale_denominator: `(cdb_XYZ_Resolution(${z}) / 0.00028)`,
+                scale_denominator: `(cdb_XYZ_Resolution(${z})::numeric / 0.00028)`,
```
rochoa: 😈

Dockerfile
dv343: No, it is not necessary, IMO before using the image in other repos cleaning was the right approach, right now... I think it is an unimportant trade-off, feel free to change it

PR #570, opened 2017-10-09
MVT buffersize fix

dv343:
This PR fix 2 buffersize problems:

- It fixes the SQL query to expand the bounding box by using the buffersize
- It fixes the pg-mvt factory to get the mapConfig buffersize value

Additionally, it fixes the query substitution for tiles with sizes different than 256 pixels.

### DIFF
--- a/lib/windshaft/renderers/pg-mvt/factory.js
+++ b/lib/windshaft/renderers/pg-mvt/factory.js
@@ -65,7 +65,9 @@ PgMvtFactory.prototype = {
 
         _.extend(dbParams, mapConfig.getLayerDatasource(layer));
         const psql = new PSQL(dbParams, this.options.dbPoolParams);
-
-        return callback(null, new Renderer(layers, psql, {}, layer));
+        if (Number.isFinite(mapConfig.getBufferSize('mvt'))) {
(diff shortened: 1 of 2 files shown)


lib/windshaft/renderers/pg-mvt/factory.js
```diff
@@ -65,7 +65,9 @@ PgMvtFactory.prototype = {
 
         _.extend(dbParams, mapConfig.getLayerDatasource(layer));
         const psql = new PSQL(dbParams, this.options.dbPoolParams);
-
-        return callback(null, new Renderer(layers, psql, {}, layer));
+        if (Number.isFinite(mapConfig.getBufferSize('mvt'))) {
+            this.options.bufferSize = mapConfig.getBufferSize('mvt');
```
rochoa: Buffer-size can be override by using CartoCSS:

```
Map {
  buffer-size: 32;
}
```

At the same time, for vector tiles, it doesn't make sense as you might not send CartoCSS because all styling/rendering will happen at the client side. So, probably, it's time to ignore it, but I wanted to let you know about that scenario.

dv343: Good point, I didn't know. But yes, I think it's better to don't offer it in the CartoCSS.

PR #573, opened 2017-11-15
Optimize Dockerfile

christianbundy:
Optimize the Dockerfile by moving `apt-get update` and `apt-get install` into the same command (to preserve image state), combining commands to reduce layers being created, and reducing both the time and size required to build and store the resulting image layers. I implemented everything I was familiar with from [the best practices](https://docs.docker.com/engine/userguide/eng-image/dockerfile_best-practices/), but I'm sure there's always room for improvement.

This removes 13 redundant layers which were taking up time and space. I'm using `set -ex` for maximum verbosity, but please let me know if you'd prefer another style. Thanks for your work on Windshaft!

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,36 +1,76 @@
 FROM ubuntu:xenial
 
 # Add external repos
-RUN apt-get -y update
-RUN apt-get install -y software-properties-common
-RUN add-apt-repository -y ppa:ubuntu-toolchain-r/test
-RUN apt-get -y update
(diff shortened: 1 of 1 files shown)


dv343:
Hi @christianbundy ,

Thank you for your PR, the `set -ex` and the layer compaction are very good.

I've fixed a problem with the locales: the locales need to be set to UTF-8 before the installation of pgsql.

I hope this change is fine for you

PR #575, opened 2017-11-28
Allow to request MVT tiles without CartoCSS

dgaubert:


dgaubert:
Ready for CR

lib/windshaft/backends/map_validator.js
```diff
@@ -2,6 +2,9 @@ var step = require('step');
 var queue = require('queue-async');
 var _ = require('underscore');
 
+const INCOMPATIBLE_LAYERS_ERROR =
+'The layergroup contains incompatible layers: don\'t mix styled layers with non styled layers (without cartocss)';
```
rochoa: What about: ```The `mapnik` or `cartodb` layers must be consistent: `cartocss` option is either present or voided in all layers. Mixing is not allowed.```

lib/windshaft/backends/map_validator.js
```diff
@@ -27,6 +30,17 @@ MapValidatorBackend.prototype.validate = function(mapConfigProvider, callback) {
 
         var token = mapConfig.id();
 
+        if (mapConfig.isVectorLayergroup()) {
+            return self.validateVectorLayergroup(mapConfigProvider, params, token, callback);
+        }
+
+        if (mapConfig.hasIncompatibleLayers()) {
+            const error = new Error(INCOMPATIBLE_LAYERS_ERROR);
+            error.type = 'layergroup';
```
rochoa: Do we use `type: 'layergroup'` in other cases?

lib/windshaft/models/mapconfig.js
```diff
@@ -123,6 +123,56 @@ function isValidBufferSize(value) {
...
+    const layers = this._cfg.layers;
+    let isVectorLayergroup = false;
+
+    if (!layers.length) {
+        return isVectorLayergroup;
+    }
+
+    isVectorLayergroup = true;
+
+    for (const layer of layers) {
```
rochoa: Shouldn't this be `let layer of layers`?

lib/windshaft/models/mapconfig.js
```diff
@@ -123,6 +123,56 @@ function isValidBufferSize(value) {
     return Number.isFinite(parseInt(value, 10));
 }
 
+MapConfig.prototype.hasIncompatibleLayers = function () {
+    return !this.isVectorLayergroup() && this.hasVectorLayer();
+};
+
+MapConfig.prototype.isVectorLayergroup = function () {
+    const layers = this._cfg.layers;
```
rochoa: If there is no go reason to use `this._cfg.layers`, you should use `this.getLayers()`.

lib/windshaft/models/mapconfig.js
```diff
@@ -123,6 +123,56 @@ function isValidBufferSize(value) {
...
+            hasVectorLayer = true;
+            break;
+        }
+    }
+
+    return hasVectorLayer;
+};
+
+MapConfig.prototype.isVectorLayer = function (layer) {
+    if (Number.isFinite(layer)) {
```
rochoa: Why do we have to support object and index param?

lib/windshaft/models/mapconfig.js
```diff
@@ -123,6 +123,56 @@ function isValidBufferSize(value) {
...
+        const layerIndex = layer;
+        layer = this.getLayer(layerIndex);
+    }
+
+    return getType(layer.type) === 'mapnik' &&
+        typeof layer.options.sql === 'string' &&
```
rochoa: What about using `getLayerOption`?

lib/windshaft/models/mapconfig.js
```diff
@@ -123,6 +123,56 @@ function isValidBufferSize(value) {
     return Number.isFinite(parseInt(value, 10));
 }
 
+MapConfig.prototype.hasIncompatibleLayers = function () {
+    return !this.isVectorLayergroup() && this.hasVectorLayer();
+};
+
+MapConfig.prototype.isVectorLayergroup = function () {
```
rochoa: What about `isVectorOnlyLayergroup()`?

lib/windshaft/renderers/mapnik/factory.js
```diff
@@ -133,6 +133,14 @@ MapnikFactory.prototype.getAdaptor = function(renderer, format, onTileErrorStrat
 
 MapnikFactory.prototype.getRenderer = function (mapConfig, format, options, callback) {
     var self = this;
+
+    if (mapConfig.isVectorLayergroup() && format !== 'mvt') {
+        const error = new Error('Invalid format: only mvt format is available for layers without CartoCSS defined');
```
rochoa: What about ```Unsupported format: `cartocss` option is missing for `${format}` format```?

lib/windshaft/renderers/pg-mvt/factory.js
```diff
@@ -44,6 +44,13 @@ PgMvtFactory.prototype = {
     },
 
     getRenderer: function (mapConfig, format, options, callback) {
+        if (mapConfig.isVectorLayergroup() && format !== 'mvt') {
+            const error = new Error('Invalid format: only mvt format is available for layers without CartoCSS defined');
```
rochoa: Something similar here.

lib/windshaft/models/mapconfig.js
dgaubert: I'm confused here with the previous comment: `Why do we have to support object and index param?`.

If in the previous comment we don't want support layer index the here we can't use `.getLayerOption(index, prop)`. What approach do you prefer? layer object or layer index?

lib/windshaft/models/mapconfig.js
dgaubert: `for of` creates a new `const` for each iteration. And we are just reading from it inside the loop, so `const` makes sense in this case, isn't it?
Further info: http://www.ecma-international.org/ecma-262/6.0/index.html#sec-for-in-and-for-of-statements-static-semantics-boundnames

lib/windshaft/backends/map_validator.js
dgaubert: Nope, but I did not want to return: 
```
{
  errors_with_context: [{
    type: 'unkown'
    message: '....'
  }]
}
```
It's basically a new type of error. It could be useful for the client, couldn't be?

rochoa: I see, but I'm not sure what is better: `layergroup` vs `mapconfig`.

lib/windshaft/models/mapconfig.js
rochoa: I didn't know that. Thank you for the clarification. 👍

lib/windshaft/models/mapconfig.js
rochoa: Hehehe. The previous comment was about not supporting both types, but if we go with the `getLayerOption` approach, you are right, you will have to use the index.

lib/windshaft/backends/map_validator.js
dgaubert: What do you like more?

lib/windshaft/models/mapconfig.js
dgaubert: perfect, going with `getLayerOption` then.

lib/windshaft/backends/map_validator.js
rochoa: I will go with `mapconfig`, the `layergroup` is what the server returns to the client, but the `mapconfig` is what the user sends, which is the root of this error.

dgaubert: cool, that's a detail that always was there but I never got it. Thanks, it has more sense for me!

dgaubert: Last question, having in mind this, should `isVectorOnlyLayergroup()` method be renamed by `isVectorOnlyMapConfig`??

rochoa: Yes.

dgaubert:
Ready for another CR round. 🤗

PR #578, opened 2017-12-11
Force png tile generation for static maps

ethervoid:
If the user tries to generate a static map requesting JPG it will fail
because is going to try to generate the tiles using JPG as format which
is not supported by now

So we now force the tiles for be generated as PNG and we ask abaculous
to generate the final image as a JPEG


Algunenano:
Tested locally and it works fine

PR #579, opened 2017-12-12
Add methods to MapConfig to support aggregation

dgaubert:


dgaubert:
I don't like this approach. I relaized that I'm relying on Windshaft's MapConfig model the whole aggreggation thing wich is the contrary that we are doing by adapting the SQL in Windshaft-CartoDB. I have a better approach for this keeping MapConfig agnostic from the aggregation. 

Stay tuned! ;)

PR #580, opened 2017-12-13
Add functions to MapConfig to discover all columns used (cartocss, interactivity and columns)

dgaubert:


dgaubert:
After changes in `windshaft-cartodb#mapconfig-aggregation` that's no longer needed!

Issue #574, opened 2017-11-28
Torque results order is unspecified


jgoizueta:
This test is now broken in some cases (apparently when using PostgreSQL 9.6 instead of 9.5):

```
  1) torque query can hold substitution tokens:
     Uncaught AssertionError: [ anonymous {
    x__uint8: 128,
    y__uint8: 128,
    vals__uint8: [ 3, 2 ],
    dates__uint16: [ 0, 1 ] } ] deepEqual [ { x__uint8: 128,
    y__uint8: 128,
    vals__uint8: [ 2, 3 ],
    dates__uint16: [ 1, 0 ] } ]
      at /home/ubuntu/windshaft/test/acceptance/torque.js:302:20
```

The order of the torque results has changed (but is consistent, note that the date to values correspondence is kept).

This is the query that returns the torque results:

https://github.com/CartoDB/Windshaft/blob/master/lib/windshaft/renderers/torque/renderer.js#L106-L138

The query doesn't specified an order, so PostgreSQL is free to return the results in any order.

We can easily force the results in the order the text expects (by date):

```diff
diff --git a/lib/windshaft/renderers/torque/renderer.js b/lib/windshaft/renderers/torque/renderer.js
index 10b502d..a9b98f6 100644
--- a/lib/windshaft/renderers/torque/renderer.js
+++ b/lib/windshaft/renderers/torque/renderer.js
@@ -133,7 +133,7 @@ Renderer.prototype = {
                 "FROM ({_sql}) i, par p " +
                 "WHERE i.{gcol} && p.b_ext " +
                 stepFilter(attrs) +
-            "GROUP BY xx, yy, d  " +
+            "GROUP BY xx, yy, d ORDER BY d desc" +
             ") cte, par  " +
             "GROUP BY x__uint8, y__uint8; ";
```

But this could affect the performance of the query.

On the other hand we're not relying on the data being ordered by date to render it on the client we could just change the test to not be order dependent and leave the query alone.

So we should check if wherever we use torque tiles we're assuming date order, or if the order doesn't affect our rendering.

ramiroaznar:
This need to be tested before the postgres 10 migration.

jgoizueta:
Hey, I've realised that I was confusing the order of the torque results:

* Our test expects the results in descending date order which seems to be the order we were getting in PostgreSQL 9.5
* With PostgreSQL 9.6 we get the data in ascending date order (which looks nicer to me).

[...]

PR #581, opened 2017-12-18
Torque tests: Sort objects before deep comparison

Algunenano:
Since the order of the torque results isn't guaranteed depending on how the query is handled by PostgreSQL we had some failing tests with PG9.6+. This PR takes care of it by ordering the objects in the tests before checking if they are equal

Tested locally with pg10. @jgoizueta Can you have a look and test if it's fixed for you too, please?

Fixes https://github.com/CartoDB/Windshaft/issues/574


dgaubert:
Hey @Algunenano, thank you so much for taking care of this issue.

The point here is that we need to know if Postgres-9.6 will break current torque visualizations. We need to create a torque layer in Builder against Postgres 9.5 and another one against 9.6 and check if there are differences.

If help is needed, let me know!

Algunenano:
So, here are my findings:
* To get the data from a provider, torque calls [getTileData](https://github.com/CartoDB/torque/blob/5d9b02a6da2c3b458fa4d60630a529b45bc5030b/lib/torque/provider/tilejson.js#L221) which calls _getTileData which is the one that does the GET and calls proccessTile before returning it. I'm using tilejson provider here but it's the same for the windshaft provider.
* The important function here is [processTile](https://github.com/CartoDB/torque/blob/5d9b02a6da2c3b458fa4d60630a529b45bc5030b/lib/torque/provider/tilejson.js#L51) which gets the rows received from the network and transforms it into a more efficient structure:
  * Checks all used dates and creates and array to hold values.
  * For each row in the received file:
    * For each date used:
       * If `!this.options.cumulative`: Adds this value to the date array (as a new node to a list).
       * If not (that is, using cumulative):  Adds an accumulated value to the date array (as a new node to a list)
  * Gathers all the results per slot in the date array and returns them as renderData and renderDataPos arrays.

Bonus: There is a bug in the [function](https://github.com/CartoDB/torque/blob/5d9b02a6da2c3b458fa4d60630a529b45bc5030b/lib/torque/provider/windshaft.js#L116) as it cannot be `this.options.cumulative` because 2 lines above it check for the opposite.

And here is when the rendering takes part and where to look depends on what you are rendering. For example for points it will end up calling [_renderTile](https://github.com/CartoDB/torque/blob/5d9b02a6da2c3b458fa4d60630a529b45bc5030b/lib/torque/renderer/point.js#L312) which will generate sprites with `generateSprite` and here it starts calling cartocss and canvas functions I don't know nothing about so I don't know if they are order independent or not. What's for sure is I haven't found anything anywhere enforcing order so, even if they where dependent on it, I don't know which order would be correct.

Algunenano:
Based on @rochoa Torque renderer test I've added another commit to prove that you get the same output independently of the order of the data, so there is no difference between PG 9.5 and PG10.

Algunenano:
Making sure the NEWs comment ends up in the current release. I'll merge if test are Ok.

PR #585, opened 2017-12-30
Support layer minzoom and maxzoom in Mapnik/CartoDB layers

rochoa:


test/acceptance/zoom-min-max.js
```diff
@@ -0,0 +1,273 @@
...
+        const suiteConfigurations = [{ mvt: { usePostGIS: false } }];
+        if (process.env.POSTGIS_VERSION === '2.4') {
+            suiteConfigurations.push({ mvt: { usePostGIS: true } });
+        }
+
+        suiteConfigurations.forEach(clientOptions => {
+            describe(`${JSON.stringify(clientOptions)}`, () => {
+
+                scenarios.forEach(scenario => {
+                    it(scenario.desc, done => {
```
dgaubert: We should be careful of using arrow functions for callbacks in mocha, it overrides the function's context for each test case in order to share objects created in before/after hooks as we do in other test suites:
```
before(function () {
  this.sut = new Wadus()
})
it('blah blah ...', () => {
  this.sut.doSomething() // Error!!!
})
```
In those tests it's fine because you aren't link anything to the test context.

rochoa: Yes, I know. Do you think it's better to keep it consistent and don't use arrow functions so it's easier to change the suite if you need the right context bound?

dgaubert: I'd say yes, better being consistent, but to be honest, these tests don't use to change at all.

rochoa:
I just changed it for the sake of consistency.

PR #587, opened 2018-01-15
Add mapnik metrics option

Algunenano:
PR to add the mapnik metrics option in Windshaft. For now set globally in the configuration file.

### DIFF
--- a/lib/windshaft/renderers/mapnik/factory.js
+++ b/lib/windshaft/renderers/mapnik/factory.js
@@ -106,7 +106,10 @@ function MapnikFactory(options) {
 
         // If enabled Mapnik will reuse the features retrieved from the database
         // instead of requesting them once per style inside a layer
-        'cache-features': true
+        'cache-features': true,
+
+        // Require stats per query to the renderer
+        metrics: false
     });
 
     this.tile_scale_factors = this._mapnik_opts.scale_factors.reduce(function(previousValue, currentValue) {
@@ -135,6 +138,16 @@ MapnikFactory.prototype.getAdaptor = function(renderer, format, onTileErrorStrat
     return new MapnikAdaptor(renderer, format, onTileErrorStrategy);
 };
 
+MapnikFactory.prototype.defineExpectedParams = function (params) {
+    if (params['cache-features'] === undefined) {
+        params['cache-features'] = this._mapnik_opts['cache-features'];
+    }
+
+    if (params.metrics === undefined) {
+        params.metrics = this._mapnik_opts.metrics;
+    }
+};
+
 MapnikFactory.prototype.getRenderer = function (mapConfig, format, options, callback) {
     var self = this;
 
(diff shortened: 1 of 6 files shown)


Algunenano:
WIP: Needs tilelive-mapnik first.

Algunenano:
Once tilelive-mapnik is merged and released, this could be merged and released. I also have ready the changes for Windshaft-Cartodb.

Algunenano:
After the release of tilelive-mapnik (0.6.18-cdb5) and some changes (mapnik_metrics -> metrics), this should be ready.

As I talked with @rafatower for now the metrics are available in the `X-Tiler-Profiler` header so, until there is a some consensus with Infrastructure and we check the impact of having it on (bigger headers) it's deactivated by default.

rafatower:
Just to be clear: it is going to be available but not enabled by default.

Not to be picky but jshint is complaining about cyclomatic complexity: https://travis-ci.org/CartoDB/Windshaft/builds/336567202#L1625

Algunenano:
Ey @rafatower / @rochoa Can I go ahead and merge this or does it need any change?

lib/windshaft/renderers/mapnik/factory.js
```diff
@@ -135,6 +138,16 @@ MapnikFactory.prototype.getAdaptor = function(renderer, format, onTileErrorStrat
     return new MapnikAdaptor(renderer, format, onTileErrorStrategy);
 };
 
+MapnikFactory.prototype.defineExpectedParams = function (params) {
```
rafatower: can't these be taken as false if not defined?

Algunenano: If it's not defined it's taken from the defaults above where it's defined as `false`.

test/acceptance/raster.js
```diff
@@ -96,5 +96,62 @@ describe('raster', function() {
...
+                        geom_column: 'rst',
+                        geom_type: 'raster',
+                        cartocss: '#layer { raster-opacity:1.0 }',
+                        cartocss_version: '2.0.1',
+                    }
+                }
+            ]
+        };
+
+        var testClient = new TestClient(mapconfig, { mapnik : { mapnik : { metrics : true } } });
```
rafatower: from my ignorance:
- how's that option passed from windshaft-cartodb?
- what's the scope and lifespan of that option?

Algunenano: From Windshaft-cartodb is set as any other Mapnik option, from the .js option file (development.js / production.js) under the `mapnik` head. You can look at [this commit](https://github.com/CartoDB/Windshaft-cartodb/commit/55fd660d695d0be3a1c7840b5f0883270b6201b0#diff-e0d6a146acab4058cec2a0311e715d6a) as an example adding `'cache-features`. `Metrics` would be just bellow it.

Lifespan is global as currently there is no specific options per renderer since there isn't any special setting per renderer/user. That is, it could be set dynamically  when calling getRenderer (as any other mapnik option) but that's not implemented in Windshaft-cartodb at the moment.

PR #592, opened 2018-01-30
Support layer 'cache-features' in Mapnik/CartoDB layers

Algunenano:
I'm kind of worry about the testing part since I'm not sure what the appropriate way to test this is (render should be the same or really similar).

If someone can have a look at the code it'd great, but feel free to ignore it.

### DIFF
--- a/test/unit/render_cache.test.js
+++ b/test/unit/render_cache.test.js
@@ -40,7 +40,8 @@ describe('render_cache', function() {
                 options: {
                     sql: 'select 2 id, null::geometry the_geom',
                     cartocss: '#layer { }',
-                    cartocss_version: '2.3.0'
+                    cartocss_version: '2.3.0',
+                    'cache-features': false
                 }
             }
         ]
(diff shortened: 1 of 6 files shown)


test/unit/render_cache.test.js
```diff
@@ -40,7 +40,8 @@ describe('render_cache', function() {
                 options: {
                     sql: 'select 2 id, null::geometry the_geom',
                     cartocss: '#layer { }',
-                    cartocss_version: '2.3.0'
+                    cartocss_version: '2.3.0',
+                    'cache-features': false
```
rafatower: Testing with this set to false seems like "cheating". This can reveal the cases where the results are not exactly the same, those we should focus on.

Algunenano: I'm only setting that one map to `false` to make sure setting it in the Mapconfig works. Any other test is using `true` or not set (which defaults to true).

Algunenano:
No, it doesn't need 3.0.14 but there are some (marginal) improvements for this case in there.

> This should be the key when it comes to how to validate them.
The easiest way to test this is to create a layer with an outline (which will have 2 styles) and compare render times between versions. If the metrics were already integrated you could also see that the datasource request are reduced from 2 to 1 and only half the time is spent there.

test/unit/render_cache.test.js
rafatower: Okay. I completely misread it, sorry.

rafatower:
> If the metrics were already integrated you could also see that the datasource request are reduced from 2 to 1 and only half the time is spent there

Are you talking about a specific PR pending integration?

Algunenano:
> Are you talking about a specific PR pending integration?
It's pending adding support in tilelive Mapnik (https://github.com/CartoDB/tilelive-mapnik/pull/10) and once that's integrated and released also here (with a similar option as this one).

rafatower:
Ok, I see now you were talking about fine-grained metrics whereas I was talking about E2E benchmarks :smile: 
I feel this is gonna be our first visible big performance step in the tiler, all thanks to you. All good from my side :+1:

PR #593, opened 2018-01-31
Do not expose 'cache-features' in the Mapconfig API

Algunenano:


test/unit/render_cache.test.js
```diff
@@ -130,13 +130,13 @@ describe('render_cache', function() {
         });
     });
 
-    it('Works with cache-features = false in layer', function(done){
+    it('Does not work with cache-features = false in layer', function(done){
```
rafatower: I'd rephrase it to "works with cache-features = true by default".

Algunenano: I'll change it to `Ignores cache-features = false in layer`. The Default is a couple of test above

rochoa:
Thank you, and sorry I didn't catch this change before.

test/unit/render_cache.test.js
Algunenano: `Ignores cache-features = false in layer (Uses default = true)'` was clearer

Algunenano:
Merging. I'll do a new release.

rafatower:
My bad I didn't catch it despite the change in the map config doc, sorry  :(

Algunenano:
Np, solved now in 4.3.3. I think this one could be merged now as it is

PR #602, opened 2018-03-06
 AttributesBackend: Allow multiple points if all the attributes are  the same 

Algunenano:
The idea is that if there are several points in the map in the same position but all have the same attributes (for the requested columns) we'll return those common attributes instead of an error. If the attributes are different we'd still return an error. 

Related to https://github.com/CartoDB/support/issues/1298

Also re-enables test cleanup that I pushed by mistake.

### DIFF
--- a/lib/windshaft/backends/attributes.js
+++ b/lib/windshaft/backends/attributes.js
@@ -84,7 +84,7 @@ AttributesBackend.prototype.getFeatureAttributes = function (mapConfigProvider, 
                 pixel_height: '1'
             });
 
-            var sql = 'select ' + quoted_att_cols + ' from ( ' + layerSql + ' ) as _windshaft_subquery ';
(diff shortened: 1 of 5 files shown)


Algunenano:
Building it's failing with yarn but works with npm. I'll try enforcing the npm modules dependencies to all use the same mapnik release

dgaubert:
oh I see, when green feel free to merge it.

lib/windshaft/backends/attributes.js
```diff
@@ -101,8 +102,8 @@ AttributesBackend.prototype.getFeatureAttributes = function (mapConfigProvider,
                     return null;
                 }
                 else {
-                    var rowsLengthError = new Error(data.rows.length + " features in layer " + params.layer +
-                        " are identified by fid " + params.fid);
+                    var rowsLengthError = new Error("Multiple features (" + data.rows.length + ") identified by '" +
```
oleurud: :heart:

PR #604, opened 2018-03-12
AttributesBackend: Support distinct and json_agg

Algunenano:
Related to https://github.com/CartoDB/support/issues/1380

Since there is no comparison operator for json objects in Postgresql, the DISTINCT with json_agg failed, so I've moved the comparisons to js instead of the query.

### DIFF
--- a/lib/windshaft/backends/attributes.js
+++ b/lib/windshaft/backends/attributes.js
@@ -84,7 +84,7 @@ AttributesBackend.prototype.getFeatureAttributes = function (mapConfigProvider, 
                 pixel_height: '1'
             });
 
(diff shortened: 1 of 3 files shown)


lib/windshaft/backends/attributes.js
```diff
@@ -102,6 +102,16 @@ AttributesBackend.prototype.getFeatureAttributes = function (mapConfigProvider,
                     return null;
                 }
                 else {
+                    if (data.rows.length > 1) {
+                        // If we receive more than one row for the id (usually `cartodb_id`)
+                        // we want to check that the attributes received are truly different before
+                        // returning an error.
+                        const uniqueAttributes = [...new Set(data.rows.map(r => JSON.stringify(r)))];
```
dgaubert: I'm not a big fan of using `JSON.stringify` to remove duplicates with `Set`, but I understand is a compromise.

PR #605, opened 2018-03-13
Fix bug when parsing incomplete Mapnik metrics

Algunenano:
Related to https://github.com/CartoDB/support/issues/1374

When partial metrics were returned from Mapnik (due to not rendering anything for example) an error appeared as it tried to read from an undefined object

### DIFF
--- a/lib/windshaft/renderers/mapnik/adaptor.js
+++ b/lib/windshaft/renderers/mapnik/adaptor.js
@@ -1,15 +1,22 @@
 var wrap = require('../base_adaptor').wrap;
 
 function parseMapnikMetrics(stats) {
+    /*jshint maxcomplexity:20 */
     if (stats && stats.hasOwnProperty('Mapnik')) {
         const setup = ((stats.Mapnik || {}).Setup || {})['Time (us)'];
-        if (setup) stats.Mk_Setup = setup / 1000;
+        if (setup) {
+            stats.Mk_Setup = setup / 1000;
+        }
 
         const render = ((stats.Mapnik || {}).Render || {})['Time (us)'];
-        if (render) stats.Mk_Render = render / 1000;
+        if (render) {
+            stats.Mk_Render = render / 1000;
+        }
 
         const features = ((((stats.Mapnik) || {}).Render || {}).Style || {}).features;
-        if (features) stats.Mk_FeatNum = features;
+        if (features) {
+            stats.Mk_FeatNum = features;
+        }
 
         delete stats.Mapnik;
     }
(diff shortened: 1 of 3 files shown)


lib/windshaft/renderers/mapnik/adaptor.js
```diff
@@ -1,10 +1,23 @@
 var wrap = require('../base_adaptor').wrap;
 
 function parseMapnikMetrics(stats) {
+    /*jshint maxcomplexity:20 */
```
dgaubert: I wouldn't disable this rule as much as possible. By extracting a function that returns, for instance, the render value:
```
function getRenderTimeFromStats(stats = {}) {
  return ((stats.Mapnik || {}).Render || {})['Time (us)'];
}
const render = getRenderTimeFromStats(stats);
```
You can reduce max complexity.

rochoa: What about default required keys before accessing them?

My proposal:

```diff
diff --git a/lib/windshaft/renderers/mapnik/adaptor.js b/lib/windshaft/renderers/mapnik/adaptor.js
index 9526ae16..ce272d6d 100644
--- a/lib/windshaft/renderers/mapnik/adaptor.js
+++ b/lib/windshaft/renderers/mapnik/adaptor.js
@@ -1,19 +1,21 @@
 var wrap = require('../base_adaptor').wrap;
 
 function parseMapnikMetrics(stats) {
-    /*jshint maxcomplexity:20 */
     if (stats && stats.hasOwnProperty('Mapnik')) {
-        const setup = ((stats.Mapnik || {}).Setup || {})['Time (us)'];
+        stats.Mapnik = Object.assign({Setup: {}, Render: {}}, stats.Mapnik);
+        stats.Mapnik.Render = Object.assign({Style: {}}, stats.Mapnik.Render);
+
+        const setup = stats.Mapnik.Setup['Time (us)'];
         if (setup) {
             stats.Mk_Setup = setup / 1000;
         }
 
-        const render = ((stats.Mapnik || {}).Render || {})['Time (us)'];
+        const render = stats.Mapnik.Render['Time (us)'];
         if (render) {
             stats.Mk_Render = render / 1000;
         }
 
-        const features = ((((stats.Mapnik) || {}).Render || {}).Style || {}).features;
+        const features = stats.Mapnik.Render.Style.features;
         if (features) {
             stats.Mk_FeatNum = features;
         }
```

Algunenano: I prefer Dani's approach (rather than set a value if not set so it can be read), but to be more consistent (and avoid future complexity issues if more metrics are added) I've used `parseXXXXXXMetric` and modified the object in each function.

rochoa: 👍

Algunenano:
@dgaubert Can this be merged now?

PR #609, opened 2018-04-04
Config for markers symbolizer caches [WIP]

rafatower:
This fully relates to https://github.com/CartoDB/grainstore/pull/145 and https://github.com/CartoDB/mapnik/pull/43

The configuration was meant to mimic the original, "extensible" proposal:
```js
var config = {
    renderer: {
        mapnik: {
            markers_symbolizer_caches: {
                disabled: false,     // markers_symbolizer_caches_disabled_
                images: {
                    size: 4096,      // markers_symbolizer_images_cache_size_
                    sampling_rate: 8 // markers_symbolizer_images_cache_sampling_rate_
                },
                attributes_size: 256, // markers_symbolizer_attrs_cache_size_  
                ellipses_size: 256,   // markers_symbolizer_ellipses_cache_size_
            }
        }
    }
};
```

As discussed, unless we instantiate the caches per-map and as long as the caches are static, the size attributes do not make sense. But the plan was to keep it flexible enough w/o introducing unneeded code/configs.

There's a mapping between Windshaft config, parameters, mml, xml and the final per-map bool config setting. There will always be such mapping. Unless we wanted to keep configs more plain, I don't see a better way to deal with defaults in the multiple layers. Yet I'm open to suggestions

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,9 +1,10 @@
-# Version 4.6.1
+# Version 4.7.0
 2018-mm-dd
 
 Announcements:
  - Adding Redis v4 in Dockerfile
- - Update @carto/mapnik to [`3.6.2-carto.6`](https://github.com/CartoDB/node-mapnik/blob/v3.6.2-carto/CHANGELOG.carto.md#362-carto6). Also update tilelive-mapnik, tilelive-bridge and abaculus accordingly.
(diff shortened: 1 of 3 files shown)


rafatower:
@Algunenano all dependencies are updated except grainstore (yet). Maybe we should split this into updating mapnik and the rest, I don't know. Failing tests are related to metrics BTW.

It is WIP but deemed good to share with you sooner rather than later.

Algunenano:
> @Algunenano all dependencies are updated except grainstore (yet). Maybe we should split this into updating mapnik and the rest, I don't know. Failing tests are related to metrics BTW.

You are right. Since the metrics changed names I think it'd best to update mapnik (and dependencies), fix the test (and add new ones if required for new metrics) in one PR and add this config (updating grainstore and adding tests) in another one.

rafatower:
Superseded by https://github.com/CartoDB/Windshaft/pull/611

PR #611, opened 2018-04-04
Config markers symbolizer caches test [WIP]

rafatower:
This is rebased on top of @Algunenano 's `master_mapnik_update` (an earlier version)

Just here to run tests.


rafatower:
And to round this up, the PR in Windshat-cartodb with the config examples: https://github.com/CartoDB/Windshaft-cartodb/pull/929/files

lib/windshaft/renderers/mapnik/factory.js
```diff
@@ -146,6 +146,10 @@ MapnikFactory.prototype.defineExpectedParams = function (params) {
     if (params.metrics === undefined) {
         params.metrics = this._mapnik_opts.metrics;
     }
+
+    if (params.markers_symbolizer_caches === undefined && this._mapnik_opts.markers_symbolizer_caches) {
```
Algunenano: I'd add the option [above](https://github.com/CartoDB/Windshaft/blob/3169002c4109eaea1467444b917a7df2539f66ac/lib/windshaft/renderers/mapnik/factory.js#L39) with a default valid for Windshaft and remove the right hand side of the `&&`.

test/acceptance/markers_symbolizer_caches_config.js
```diff
@@ -0,0 +1,90 @@
...
+                    }
+                }
+            }
+        };
+
+        var testClient = new TestClient(MAPCONFIG, options);
+        testClient.getTile(0, 0, 0, {format: FORMAT}, function(err, tile, img, headers, stats) {
+            assert.ok(!err);
+            assert.equal(stats.hasOwnProperty('Mk_Agg_PMS_ImageCache_Miss'), true);
+            assert.equal(stats.hasOwnProperty('Mk_Agg_PMS_ImageCache_Ignored'), false);
```
Algunenano: Maybe adding an extra point to the dataset and checking that we get a `Mk_Agg_PMS_ImageCache_Hit` is more appropriate since the `ignored` can also be triggered depending on the style.

Algunenano: I see now that the `Mk_Agg_PMS_ImageCache_Miss` value changes too so there is no need. I don't know why I thought only the `ignored` was changing.

lib/windshaft/renderers/mapnik/factory.js
Algunenano: Here is how I see it. Mapnik should have a default (in Mapnik) and it shouldn't be setup anywhere else in the stack, but for my point of view since Windshaft is the end of the stack (or can be, since we have Windshaft-cartodb on top) it should have its appropriate defaults. We did it a similar with metrics.

In fact I'd also add this new option in `config/environments/development.js` for visibility.

Algunenano: Also keep in mind that setting the default above avoids any need to check if `this._mapnik_opts.markers_symbolizer_caches` is defined or not.

rafatower: let's talk to settle this, please

test/acceptance/markers_symbolizer_caches_config.js
rafatower: :+1:

lib/windshaft/renderers/mapnik/factory.js
rafatower: Addressed in f4ba901

rafatower:
@dgaubert :pray: small review, we're almost done!

PR #612, opened 2018-04-09
Add Mapnik metrics docs and extra tests

Algunenano:
Extra notes:
- Adds a disabled test for metatiles. Pending fix in tilelive-mapnik to enable it.
- Adds a disabled test with render time variables in the `sql`. Pending fix in tilelive-mapnik to enable it.
- Adds a disabled test for the Group Symbolizer. The `carto` module currently doesn't support, but I prefer to have it there for reference.
- The symbolizer tests could do something fancier than:
`assert(stats.hasOwnProperty('Mk_Agg_PGroupS') || stats.hasOwnProperty('Mk_Grid_PGroupS'));`,
like `assert(stats.hasOwnProperty('Mk_' + renderer + '_PGroupS') `, but I prefer this way as it leaves the whole metric name intact in case someone greps the source files looking for it.

### DIFF
--- a/test/acceptance/metrics.js
+++ b/test/acceptance/metrics.js
@@ -2,6 +2,8 @@ require('../support/test_helper');
 
 var assert = require('../support/assert');
 var TestClient = require('../support/test_client');
+var fs = require('fs');
+var http = require('http');
 
 describe('metrics', function() {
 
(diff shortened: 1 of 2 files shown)


test/acceptance/metrics.js
```diff
@@ -126,77 +128,107 @@ describe('metrics', function() {
     });
     });
 
-    it("Geometry counts", function(done) {
+    //TODO: Pending fix in tilelive-mapnik
+    it.skip("works with metatiles", function(done) {
```
rafatower: for the record: https://github.com/CartoDB/tilelive-mapnik/pull/18

PR #615, opened 2018-04-12
Add test with mvt bug

Algunenano:
Test reproducing the interaction between `false` and integer columns

~~Weirdly enough, `TRUE` seems to be always transformed into `1`, never `true`.~~

cc/ @oleurud @rochoa


Algunenano:
Should be fixed now.

PR #618, opened 2018-05-03
Using new docker image

oleurud:
Due to https://github.com/CartoDB/Windshaft-cartodb/pull/945 and https://github.com/CartoDB/CartoDB-SQL-API/pull/482 we have created a new Docker image with Xenial and Postgres 10.1

The objective of the PR:
- use the same docker image
- remove old Dockerfile

The Dockerfile has been moved to Windshaft-carto (`/docker` folder)


.travis.yml
```diff
@@ -1,13 +1,12 @@
 sudo: required
-dist: trusty
 
 services:
   - docker
 
 before_install:
-  - docker pull cartoimages/windshaft-testing
+  - docker pull cartoimages/engine-xenial-pg101
```
davidmanzanares: I don't know if this has already been decided elsewhere, but I would expect to have here the dockerfile for testing this. From my point of view, "engine" is a Carto concept, not a Windshaft concept, and in my opinion, Windshaft should be self-contained, particularly because it is an open source project.

oleurud: I agree with `engine` vs `windshaft` concept.  But the image is used by Windshaft, Windshaft-carto and CartoDB-SQL-API (not only Windshafts, also SQL API), and this was the reason for the change.

Do you find another name? I don't for now.

oleurud: `nodejs-xenial-pg101`?

davidmanzanares: Well... I would have expected that the other repos reused this one to be honest

PR #623, opened 2018-05-21
Upgrade dependencies version

dgaubert:


Algunenano:
Some notes:
- `queue-async` is deprecated. Should we use `d3-queue` instead or do you prefer keeping the old package?
- `tilelive` last release is 6.0. Any reason to not use that?
- `mocha` last release is 5.2.0. I guess there might be some issue with not automatically force closing the test if they leave open sockets, right? The old behaviour can be reproduced using `--exit` (or by fixing the tests). 

## Security (building Windshaft-cartodb):
#### Pre-update:
```
[!] 49 vulnerabilities found [2253 packages audited]
    Severity: 10 Low | 33 Moderate | 5 High | 1 Critical
    Run `npm audit` for more detail
```

#### After update:
```
[!] 49 vulnerabilities found [2259 packages audited]
    Severity: 11 Low | 32 Moderate | 5 High | 1 Critical
    Run `npm audit` for more detail
```

## Performance 
Using random point datasets [0, 10, 100, 1000, 10000, 100000], marker symbolizer, 30 warm-up requests, 3000 requests analyzed):
[pre_update_large_vs_post_update_large.pdf](https://github.com/CartoDB/Windshaft/files/2022007/pre_update_large_vs_post_update_large.pdf)
I don't see any significant difference (a slight improvement if anything at all).

dgaubert:
> queue-async is deprecated. Should we use d3-queue instead or do you prefer keeping the old package?

Moving to `d3-queue` requires changes in the code base, there are some breaking changes that must be resolved and I didn't want to add more entropy to this PR. I want to tackle this in a separated PR.

> tilelive last release is 6.0. Any reason to not use that?

`yarn outdated` didn't tell me anything about `tilelive@6.0` ¯\_(ツ)_/¯

> mocha last release is 5.2.0. I guess there might be some issue with not automatically force closing the test if they leave open sockets, right? The old behaviour can be reproduced using --exit (or by fixing the tests).

As mentioned above, I want to tackle this in a separated PR.

Algunenano:
I think that yarn might have failed to warn you about tilelive because it changed scope from global to `@mapbox`.

dgaubert:
> I think that yarn might have failed to warn you about tilelive because it changed the scope from global to @mapbox.

Yep, that's it. As this requires changes in the codebase, I'll tackle this in a separated PR

Issue #629, opened 2018-06-20
Failing test: toque_zero_zero.js


Algunenano:
The torque_zero_zero.js test is failing in my setup (Postgis 2.5 trunk, proj 5.1) where the point (0,0) is contained in all 4 tiles and not just the bottom ones:

```
$ ./node_modules/.bin/mocha -u tdd -t 5000 test/acceptance/torque_zero_zero.js


  torque tiles at 0,0 point
    1) tl
    2) tr
    ✓ bl
    ✓ br


  2 passing (63ms)
  2 failing

  1) torque tiles at 0,0 point tl:

      Uncaught AssertionError [ERR_ASSERTION]: [ anonymous {
    x__uint8: 1,
    y__uint8: 0,
    vals__uint8: [ 1 ],
    dates__uint16: [ 0 ] } ] deepEqual []
      + expected - actual

      -[
      -  {
      -    "dates__uint16": [
      -      0
      -    ]
      -    "vals__uint8": [
      -      1
      -    ]
      -    "x__uint8": 1
      -    "y__uint8": 0
      -  }
      -]
      +[]
      
      at test/acceptance/torque_zero_zero.js:83:24
      at test/support/test_client.js:87:16
      at Function.<anonymous> (lib/windshaft/backends/tile.js:61:13)
      at next (node_modules/step/lib/step.js:51:23)
      at lib/windshaft/renderers/torque/renderer.js:171:17
      at ctor.PSQLAdaptor.ctor._handleResult (lib/windshaft/renderers/torque/psql_adaptor.js:14:9)
      at Result.<anonymous> (node_modules/cartodb-psql/lib/psql.js:302:24)
      at Result.Query.handleReadyForQuery (node_modules/cartodb-psql/node_modules/pg/lib/query.js:144:10)
      at Connection.<anonymous> (node_modules/cartodb-psql/node_modules/pg/lib/client.js:184:19)
      at Socket.<anonymous> (node_modules/cartodb-psql/node_modules/pg/lib/connection.js:135:12)
      at addChunk (_stream_readable.js:274:12)
      at readableAddChunk (_stream_readable.js:261:11)
      at Socket.Readable.push (_stream_readable.js:218:10)
      at TCP.onread (net.js:581:20)

  2) torque tiles at 0,0 point tr:

      Uncaught AssertionError [ERR_ASSERTION]: [ anonymous {
    x__uint8: 0,
    y__uint8: 0,
    vals__uint8: [ 1 ],
[...]

PR #630, opened 2018-06-20
Torque 0,0 test: Remove srid transformation

Algunenano:
Attempt to fix https://github.com/CartoDB/Windshaft/issues/629 so it doesn't depend on the proj release


Algunenano:
@jgoizueta This looks right up your alley, can you have a quick look please?

Issue #628, opened 2018-06-19
Uniformize the columns returned in a MVT for Mapnik and St_AsMVT


Algunenano:
If I instantiate a map through Builder and request an MVT file by just replacing the extension, `.png` with `.mvt`, I get 2 different files depending whether `mvt::usePostGIS` is enabled or not. With it disabled, Mapnik is used and all the columns in the map query are returned. With it enabled only `the_geom_webmercator` is returned.

Since I understand we want to be able to switch between the 2 of them flawlessly this is certainly a major issue, as anything that is currently using the MVT in production and relying on other columns to be returned will be broken once the direct ST_AsMVT is enabled.

I'm proposing changing the default for ST_AsMVT to return all columns in the tile query to match the current behaviour.  @davidmanzanares @rochoa , since I see you were the ones that implemented it, what do you think?

Algunenano:
WIP in https://github.com/Algunenano/Windshaft/commit/0fa06ca84c55bd1e9c59bcd55329addf21f1ef64

rochoa:
I want to think about it.

My current memory dump:
1. Sending all the columns is inefficient.
1. Breaking the existing behaviour is unacceptable
1. We could send all columns depending on the MapConfig's options, e.g. if cartocss is present, we send all columns for MVT tiles.
1. We could send all columns always, except if the MapConfig has a `columns` option requesting only specific columns.
1. We could create a `/api/v2/map` endpoint where we change the existing behaviour, not breaking existing clients.

davidmanzanares:
I agree with @rochoa, but I would add:

6. Users of CARTO VL won't need this feature on Windshaft to get the same result (we tweak the SQL query to just get the required columns).
7. Developers that know the cost of downloading all columns can get the same result by changing the SQL query.
8. Supporting this extra feature is costly.

So, in my opinion, storing all columns is the better option since I don't think its advantages worth the cost.

Algunenano:
> Sending all the columns is inefficient.

It's what we are currently doing with mapnik. If the pg-mvt renderer only returns the geometry by default, any map currently using mvt's and expecting the columns to be included will break once we switch `usePostgis` to on.

> Breaking the existing behaviour is unacceptable

I haven't any usage of `layer.options.columns` anywhere except Windshaft own tests, not even in Windshaft-cartodb. Also it wasn't documented.
Windshaft-cartodb has some examples with `layer.options.attributes.columns` but that's intended for attributes backend, not mvts.

[...]

PR #632, opened 2018-07-02
Uniform mvt

Algunenano:
This PR tries to uniformize the MVT interface in Windshaft, with the intention of having the same output using either Mapnik or pg-mvt renderers. As Mapnik's MVTs are already in use it modifies pg-mvt to match its behaviour.

Main changes:
- pg-mvt: Drops the `layer.option.columns` to request only certain columns from the query. It was unused and not implemented in Mapnik's side.
- pg-mvt: Only requests columns of known types in Mapnik (bool, int, float, text).
- pg-mvt: Now supports `geom_column` properly (before it was only working with `the_geom_webmercator` .
- pg-mvt: Now returns an error when the tile is empty (to match Mapnik behaviour)
- Tests: MVT tests are now ignored (instead of omitted) when st_asmvt isn't supported.
- Tests: I've added a new test suit that compares the output of Mapnik and Postgis given the same query. Any test containing the `known_issue` is ignored as the behaviour between the 2 renderers is different.

This is working fine with postgis trunk and I expect it to fail with our 2.4 branch. I don't intend to merge this until the tests pass with out Postgis release, but please have a look at the PR anyway.

Thanks to @enekid and specially @IagoLast for their help dealing with my JS lack of knowledge.

cc/ @davidmanzanares and @rochoa since we were discussing some of the changes in https://github.com/CartoDB/Windshaft/issues/628


lib/windshaft/renderers/pg-mvt/renderer.js
```diff
@@ -29,70 +38,123 @@ module.exports = class PostgresVectorRenderer {
...
-                );
-                const query = this._replaceTokens(layer.options.sql, {z, x, y});
-
-                queries.push(this._vectorLayerQuery(layer.id, columns, query, { z, x, y }));
+    _getLayerColumns (layer) {
+        return new Promise((resolve, reject) => {
+            if (layer._metadata_columns) {
+                resolve(layer._metadata_columns);
             }
+            var layer_sql = this._replaceTokens(layer.options.sql, 0, 0, 0);
```
enekid: I would use a `const` and the name of the variable should adopt `javascript` conventiones:

`
const layerSQL = this._replaceTokens(layer.options.sql, 0, 0, 0);
`

lib/windshaft/renderers/pg-mvt/renderer.js
```diff
@@ -29,70 +38,123 @@ module.exports = class PostgresVectorRenderer {
...
+            var limitedQuery = `SELECT * FROM (` + layer_sql + `) __windshaft_mvt_schema LIMIT 0;`;
```
enekid: If you use template literals (\`\`), you can use `${}` instead of the `+`s

`
const limitedQuery = `SELECT * FROM (${layer_sql}) __windshaft_mvt_schema LIMIT 0;`;
`

lib/windshaft/renderers/pg-mvt/renderer.js
```diff
@@ -25,74 +34,127 @@ module.exports = class PostgresVectorRenderer {
...
-            const headers = { 'Content-Type': 'application/x-protobuf' };
-            const stats = timer.getTimes();
+                const headers = { 'Content-Type': 'application/x-protobuf' };
+                const stats = timer.getTimes();
 
-            return callback(null, result[0].st_asmvt, headers, stats);
-        });
+                return callback(null, mvt, headers, stats);
+            });
+        }, err => { return callback(err); })
```
dgaubert: To prevent handling error case in two different places, do as follows:

```
Promise.all([ ... ])
  .then(res => ...)
  .catch(err => callback(err))
```
Any error will be caught by `.catch` either the error came from `.all` or from `.then`

lib/windshaft/renderers/pg-mvt/renderer.js
```diff
@@ -25,74 +34,127 @@ module.exports = class PostgresVectorRenderer {
...
+        .catch(() => { }); // Avoids UnhandledPromiseRejectionWarning
```
dgaubert: If something went wrong in `.all` then you are hiding the underlying error. See the comment above. ^^^

lib/windshaft/renderers/pg-mvt/renderer.js
```diff
@@ -25,74 +34,127 @@ module.exports = class PostgresVectorRenderer {
...
+                    }
+                    return callback(err);
+                }
 
-            if (result.length <= 0 || result[0].st_asmvt === undefined) {
-                return callback(new Error(`Couldn't generate tile`));
-            }
+                const mvt = extractMVT(data);
+                if (!mvt) {
+                    return callback(new Error(empty_tile_msg));
```
dgaubert: I understand you are trying to unify the behaviour of both `mapnik` and `pg_asmvt` renderers. `Tilelive bridge` is the one that returns the very same error when the MVT is empty for a given set of coordinates then in `Windshaft-cartodb` we respond with `204 No Content`, among other things, to be able to cache that response at `CDN` level. Thanks to that I realized that we are doing that almost right. The fact of returning an error here makes that we are skipping the generation of cache headers (surrogate headers, cache-channel, etc..). 

So we have to options either modify `tilelive-bridge` to not return an error when the MVT is empty and, also implement the same here or handle this error in [`windshaft-cartodb`](https://github.com/CartoDB/Windshaft-cartodb/blob/028c3f9aeca6f39ad8736fc7d46724b51410d2ab/lib/cartodb/api/map/tile-layergroup-controller.js#L108-L110) to continue the normal req/res cycle generating the headers properly.

I leave to you that decission, we will have to tackle this sooner than later. If you consider that's out of the scope of thid PR, please open new issue.

Algunenano: I think the proper solution is to return a `200` with an empty tile mainly because that removes the need of parsing the error message and the special case. Since it involves modifying other modules I'll create a ticket here to adapt both renderers.

Algunenano: Done in https://github.com/CartoDB/Windshaft/issues/633. I'll tackle it in the near future

dgaubert: thanks!

test/acceptance/mvt.js
```diff
@@ -481,61 +372,604 @@ function mvtTest(usePostGIS) {
...
+            const testClientPg_mvt = new TestClient(mapConfig, { mvt: { usePostGIS: true } });
+            const options = { format : 'mvt' };
+            const z = test.tile && test.tile.z ? test.tile.z : 0;
+            const x = test.tile && test.tile.x ? test.tile.x : 0;
+            const y = test.tile && test.tile.y ? test.tile.y : 0;
+
+            testClientMapnik.getTile(z, x, y, options, function (err1, mapnikMVT) {
+                testClientPg_mvt.getTile(z, x, y, options, function (err2, pgMVT) {
+                    if (err1 || err2) {
+                        assert.ok(err1, "Mapnik didn't fail. Postgis error: " + err1);
```
dgaubert: What happens if we receive an error here but is not the error what we expect? We should separate the tests that check that everything went ok from the tests that something went wrong, and assert the expectations accordingly.

test/acceptance/mvt.js
```diff
@@ -481,61 +372,604 @@ function mvtTest(usePostGIS) {
...
+        it(test.name, function (done) {
+
+            const testClientMapnik = new TestClient(test.mapConfig, { mvt: { usePostGIS: false } });
+            const testClientPg_mvt = new TestClient(test.mapConfig, { mvt: { usePostGIS: true } });
+            const options = { format : 'mvt' };
+
+            testClientMapnik.getTile(test.tile.z, test.tile.x, test.tile.y, options, function (err1, mapnikMVT) {
+                testClientPg_mvt.getTile(test.tile.z, test.tile.x, test.tile.y, options, function (err2, pgMVT) {
+                    if (err1 || err2) {
+                        assert.ok(err1);
```
dgaubert: Shame here. We should separate error cases from the others.

lib/windshaft/renderers/pg-mvt/renderer.js
```diff
@@ -25,74 +34,127 @@ module.exports = class PostgresVectorRenderer {
...
+            const limitedQuery = `SELECT * FROM (${layerSQL}) __windshaft_mvt_schema LIMIT 0;`;
+
+            this.psql.query(limitedQuery, function (err, data) {
+                if (err) {
+                    if (err.message) {
+                        err.message = "PgMvtRenderer: " + err.message;
+                    }
+                    return reject(err);
+                }
+                else {
```
dgaubert: Unnecessary `else`.

test/acceptance/mvt.js
Algunenano: Maybe using the `expected_error` more extensively? I mean, declaring it in all the tests that are expected to fail as with the `Null geometry` test case:
```
        {
            name: 'Null geometry',
            sql: "SELECT 2 AS cartodb_id, null as the_geom",
            expected_error : "Error: Tile does not exist"
        },
```

Or would you rather have them all separated by expected to fail and expected to be ok?

dgaubert: Both suggestions are ok. It might that having a separated error case test suite will improve the readability. Don't you think?

Algunenano: I changed to check for the error message for now but, as all of them are errors due to an empty tile and that's gonna be changed, I'll revisit it when that's changed.

lib/windshaft/renderers/pg-mvt/renderer.js
```diff
@@ -25,74 +34,124 @@ module.exports = class PostgresVectorRenderer {
...
+                    const query = this._replaceTokens(layer.options.sql, { z, x, y });
+                    return this._vectorLayerQuery(layer, columns, query, { z, x, y });
+                });
+        });
     }
 
-    _geomColumn ({ z, x, y }) {
+    _geomColumn (layer, { z, x, y }) {
         // Geometry column
         const subpixelBufferSize = Math.round(this.buffer_size * this.mvt_extent / this.tile_size);
```
oleurud: Only a comment. It is not related to your changes, but we could calculate `subpixelBufferSize` once in the constructor.

Algunenano: Nice catch. I also see now that I'm not caching the column names although I check if it's cached (:S). I'll fix that too.

lib/windshaft/renderers/pg-mvt/renderer.js
```diff
@@ -25,74 +34,125 @@ module.exports = class PostgresVectorRenderer {
...
-                const geomColumn = this._geomColumn({z, x, y});
-                const columns = removeDuplicates(
-                    [ geomColumn ].concat(LayerColumns.getColumns(layer.options))
-                );
-                const query = this._replaceTokens(layer.options.sql, {z, x, y});
-
-                queries.push(this._vectorLayerQuery(layer.id, columns, query, { z, x, y }));
+    _getLayerColumns (layer) {
+        return new Promise((resolve, reject) => {
+            if (layer._mvtColumns) {
```
rochoa: This smells. If this is we do just once, it should be in the factory, populated once and injected into the Renderer.

Algunenano:
This is green now with the new 2.4.4.2+carto-1 Postgis release (Thanks @oleurud and @enekid). Do any of you want to have a quick review after all the changes?

The change to not throw an error with empty MVT will come in a different PR.

lib/windshaft/backends/tile.js
```diff
@@ -22,43 +18,41 @@ module.exports = TileBackend;
...
+        if (is_cached) {
+            extraHeaders['X-Windshaft-Cache'] = Date.now() - renderer.ctime;
+        }
 
-    var renderer;
+        timer.start('render-' + params.format.replace('.','-'));
 
-    timer.start('getTileOrGrid');
-    var extraHeaders = {};
-    step(
```
oleurud: :muscle:

lib/windshaft/backends/tile.js
```diff
@@ -22,43 +18,41 @@ module.exports = TileBackend;
 // * cache_buster - specify to ensure a new renderer is used
 // * geom_type - specify default style to use if no style present
 TileBackend.prototype.getTile = function(mapConfigProvider, params, callback) {
+    if (params.format === 'grid.json' && !params.interactivity) {
+        if ( ! params.token ) { // token embeds interactivity
+            callback(new Error("Missing interactivity parameter"));
```
enekid: aren't you missing here a `return`?
```javascript
return callback(new Error("Missing interactivity parameter"));
```

lib/windshaft/renderers/pg-mvt/factory.js
```diff
@@ -20,6 +21,74 @@ function PgMvtFactory(options) {
...
+                700,    // float4
+                701,    // float8
+                1700,   // numeric
+                1042,   // bpchar
+                1043,   // varchar
+                25,     // text
+                705,    // literal
+            ];
+
+            let metadataColumns = [];
```
enekid: better use `const`:
```javascript
const metadataColumns = [];
```

lib/windshaft/renderers/pg-mvt/renderer.js
```diff
@@ -54,48 +66,40 @@ module.exports = class PostgresVectorRenderer {
...
-                    [ geomColumn ].concat(LayerColumns.getColumns(layer.options))
-                );
-                const query = this._replaceTokens(layer.options.sql, {z, x, y});
-
-                queries.push(this._vectorLayerQuery(layer.id, columns, query, { z, x, y }));
-            }
-
-            return queries;
-        }, []);
+        return this.layers.filter(l => shouldUseLayer(l, z)).map(layer => {
```
enekid: use `layer`  instead of `l` and IMO spread it across multiple lines
```javascript
return this.layers
                  .filter(layer => shouldUseLayer(layer, z))
                  .map(layer => {
```

lib/windshaft/backends/tile.js
Algunenano: You are absolutely right

lib/windshaft/renderers/pg-mvt/factory.js
Algunenano: I thought `const` was equivalent the c++ const (does not change), but I see it means `won’t be reassigned`. I'll change it.

PR #637, opened 2018-07-12
Validate buffer size

dgaubert:


lib/windshaft/models/mapconfig.js
```diff
@@ -106,7 +114,7 @@ function getType(type) {
 }
 
 MapConfig.prototype.setBufferSize = function (bufferSize) {
-    this._cfg.bufferSize = bufferSize;
+    this._cfg.buffersize = bufferSize;
```
jgoizueta: good catch. we don't seem to be using that method, though

Issue #642, opened 2018-07-18
Review and fix attribute validation


jgoizueta:
See #639. This problem of slow attribute validation with aggregations required some quick fix, but we should review the issue again to see if a better solution is needed.

Issue #639, opened 2018-07-17
Slow validation for aggregations


jgoizueta:
In some cases of large tables using aggregation, the validation performed during map instantiation can be very slow.

The slow part of validation is [`validateAttributes`] (https://github.com/CartoDB/Windshaft/blob/f2e3535ab06906cb6eb6acc7bdb3495b3105f249/lib/windshaft/backends/map_validator.js#L62-L63) which uses [getFeatureAttributes](https://github.com/CartoDB/Windshaft/blob/f2e3535ab06906cb6eb6acc7bdb3495b3105f249/lib/windshaft/backends/attributes.js#L32-L33)

The problem is that `getFeatureAttributes` executes the query substituting `!bbox!` by the whole world limits (to make sure that if bbox is used to filter the query not all features are excluded) and with a `LIMIT 1`.  For aggregations, which use `!bbox!`, this means that the whole table has to be aggregated before one result row is taken. For large tables with millions of points this can take more than a minute easily.

As discussed with @dgaubert who debugged this problem, we could treat the aggregation case diffferently and skip this particular validation. We could optimize this case by using the query without aggregation with `LIMIT 1` to fetch one geometry and then generate a small bbox that contains it, finally using that bbox and LIMIT 1 with the aggregated query, but it's probably not worth the effort.

PR #643, opened 2018-07-18
Skip attribute validation when sql_raw is present

jgoizueta:
This is a quick fix for #639

sql_raw presence indicates aggregation or other complex query wrapping is in place.
In these cases running a query with a whole-earth bbox as the attribute validation does can be very expensive (even with LIMIT 1, because the whole table must first be aggregated).

So here we avoid the problematic attribute validation altogether when sql_raw is present.
In particular, notice that:
* If no aggregations or other sql wrapping occurs, no sql_raw is present and validation is not skipped
* If aggregations are requested but they are not being applied (too few rows or wrong geometry type) sql_raw is also empty and validation is not skipped

We'll work on a better solution in #642

### DIFF
--- a/lib/windshaft/backends/map_validator.js
+++ b/lib/windshaft/backends/map_validator.js
@@ -122,7 +122,8 @@ MapValidatorBackend.prototype.validate = function(mapConfigProvider, callback) {
                     }
 
                     // both 'cartodb' or 'torque' types can have attributes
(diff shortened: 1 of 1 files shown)


jgoizueta:
Hey guys, I've been testing this in staging and it seems to work fine, So I'll release it

Issue #644, opened 2018-08-02
Remove 'step'


oleurud:
Following the https://github.com/CartoDB/Windshaft-cartodb/issues/1010 issue, we were removing step from some modules we noticed that some metrics were improving. Let's remove step to handle async code in favor of Node.js callback pattern as long as possible.

Let's go module by module with its own PR:
- [ ] lib/windshaft/backends/attributes.js
- [ ] lib/windshaft/backends/map_validator.js
- [ ] lib/windshaft/backends/preview.js
- [ ] lib/windshaft/metadata/torque-layer-metadata.js
- [ ] lib/windshaft/renderers/mapnik/factory.js
- [ ] lib/windshaft/renderers/plain/image_renderer.js
- [ ] lib/windshaft/renderers/renderer_factory.js
- [ ] lib/windshaft/renderers/torque/factory.js
- [ ] lib/windshaft/storages/mapstore.js

oleurud:
For each Windshaft PR, we should create a new one in Windshaft-carto. Branch by branch, deploy it first in 1 server, check for no errors and then merge and deploy for every server.

oleurud:
This task is blocked by `pg-mvt` work https://github.com/CartoDB/Windshaft/pull/646 

`Windshaft` has changes that need to make some integration code in `Windshaft-carto` (mainly in test). After talking with @Algunenano, it will be better to deploy a new version with `pg-mvt` and then, continue with this issue (this one doesn't have any priority)

oleurud:
I remove hte blocked because I can continue with the acceptance, leaving everything ready for the deploy.

oleurud:
## Aceptance
- [ ] [windshaft-remove-step-attributes](https://github.com/CartoDB/Windshaft-cartodb/compare/master...windshaft-remove-step-attributes) lib/windshaft/backends/attributes.js 
- [ ] [windshaft-remove-step-map-validator](https://github.com/CartoDB/Windshaft-cartodb/compare/master...windshaft-remove-step-map-validator) lib/windshaft/backends/map_validator.js
- [ ] [windshaft-remove-step-preview](https://github.com/CartoDB/Windshaft-cartodb/compare/master...windshaft-remove-step-preview) lib/windshaft/backends/preview.js
[...]

PR #645, opened 2018-08-02
Remove step attributes

oleurud:


oleurud:
For each Windshaft PR, we should create a new one in Windshaft-carto and deploy them one by one (first in 1 server, and then to all servers)

dgaubert:
ok, 🚀

oleurud:
@dgaubert  I have updated the way to make the acceptance and deploy:
https://github.com/CartoDB/Windshaft/issues/644#issuecomment-416475851

Issue #641, opened 2018-07-17
[Mapnik MVT] Simplify based on the extent instead of the default tile-size


Algunenano:
Currently Mapnik's MVT is simplifying geometries based on the tile size, which isn't set and defaults to 256. 

When you request an MVT with extent 4096 you do so because you expect to be able to overzoom 4 levels and have the ability to have a precision of 4 cms at zoom 18 ([Explanation here](https://github.com/mapbox/vector-tiles/issues/45#issuecomment-325781516)). If you are simplifying supposing the geometries are going to be rendered in a raster of 256 pixels (default) you are breaking the overzoom capability as the geometries are simplified 255x more than necessary for the 4 overzoom levels.

Current implementation of Mapnik's MVTs using tilelive-bridge doesn't set any defaults, so the tile size is set to 256 and the MVT extent to 4096, which causes the issue mentioned above. The solution would be to set the tile-size to the same value as the extent when requesting [the tile](https://github.com/CartoDB/tilelive-bridge/blob/5129e43223cb55daed31373c7a36c98eb6178fc1/index.js#L271):
```js
    map.extent = vtile.extent();
    map.height = vtile.tileSize;
    map.width = vtile.tileSize;
```

Also, it'd nice if the simplification parameters when calling `ST_Simplify(ST_RemoveRepeatedPoints` were the same as the ones used by Postgis, but that's a minor issue.

This is needed in order to implement https://github.com/CartoDB/Windshaft-cartodb/issues/843 with compatibility in both pg-mvt and mapnik, so since the idea is to make the extent configurable (either when requesting the tile or when instantiating the map) we should keep it in mind and do both if possible.

cc/ @davidmanzanares for reference

Issue #633, opened 2018-07-03
MVT: Do not return an error with empty tile


Algunenano:
Currently if the MVT is empty we are returning an error to Windshaft-cartodb that then needs to handle this special case to return a 204 with the proper headers so it gets cached (check [this comment](https://github.com/CartoDB/Windshaft/pull/632#discussion_r199577177)).

Should be handled the same way for both Mapnik and pg-mvt renderers.

PR #646, opened 2018-08-03
Update @carto/mapnik to 3.6.2-carto.11

Algunenano:
- Dependency update from mapnik renderer stack:
  - node-mapnik 3.6.2-carto.11: Change the MVT simplification tolerance from tile_width / 256 to tile_width / tile_extent (default 4096).
  - carto/tilelive-bridge 2.5.1-cdb10: Remove error on empty MVT. Now returns an empty buffer and, as before, `x-tilelive-contains-data` set to false.
  - abaculus to 2.0.3-cdb11: Keep up with node-mapnik update.
  - tilelive-mapnik to 0.6.18-cdb15: Rework to remove step, evenEmitter and nextTick.
- MVTs no longer return an error when empty.

~~Pending for this PR:~~
- [x] Fixing tests related to empty MVTs.
- [x] NEWs.
- [x] Performance report for tilelive-mapnik changes impact (empty or low geometry tiles [png, grid.json], with and without metatile).
- [x] Performance report for MVTs after simplification change.

Closes https://github.com/CartoDB/Windshaft/issues/641
Closes https://github.com/CartoDB/Windshaft/issues/633

### DIFF
--- a/lib/windshaft/renderers/pg-mvt/renderer.js
+++ b/lib/windshaft/renderers/pg-mvt/renderer.js
@@ -45,11 +45,10 @@ module.exports = class PostgresVectorRenderer {
     /// @param z tile zoom
     /// callback: will be called when done using nodejs protocol (err, data)
     getTile (z, x, y, callback) {
-        const emptyTileMsg = 'Tile does not exist';
         const subqueries = this._getLayerQueries({ z, x, y });
 
         if (subqueries.length === 0) {
-            return callback(new Error(emptyTileMsg));
+            return callback(null, new Buffer(0));
         }
 
         const query = `SELECT ((${subqueries.join(') || (')})) AS st_asmvt`;
@@ -67,14 +66,13 @@ module.exports = class PostgresVectorRenderer {
             }
 
             const mvt = extractMVT(data);
-            if (!mvt) {
-                return callback(new Error(emptyTileMsg));
-            }
-
-            const headers = { 'Content-Type': 'application/x-protobuf' };
+            const headers = {
+                'Content-Type': 'application/x-protobuf',
+                'x-tilelive-contains-data' : mvt ? true : false
+            };
             const stats = timer.getTimes();
 
-            return callback(null, mvt, headers, stats);
+            return callback(null, mvt || new Buffer(0), headers, stats);
         });
     }
 
--- a/lib/windshaft/renderers/pg-mvt/renderer.js
+++ b/lib/windshaft/renderers/pg-mvt/renderer.js
@@ -45,11 +45,16 @@ module.exports = class PostgresVectorRenderer {
     /// @param z tile zoom
     /// callback: will be called when done using nodejs protocol (err, data)
     getTile (z, x, y, callback) {
-        const subqueries = this._getLayerQueries({ z, x, y });
+        const headers = {
+            'Content-Type': 'application/x-protobuf',
+            'x-tilelive-contains-data' : false
+        };
 
+        const subqueries = this._getLayerQueries({ z, x, y });
         if (subqueries.length === 0) {
-            return callback(null, new Buffer(0));
+            return callback(null, new Buffer(0), headers);
         }
+
 
         const query = `SELECT ((${subqueries.join(') || (')})) AS st_asmvt`;
         const timer = new Timer();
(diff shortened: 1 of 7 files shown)


Algunenano:
No visible performance impact due to `tilelive-mapnik` changes: [11_585_2_vs_11_586_2.pdf](https://github.com/CartoDB/Windshaft/files/2256815/11_585_2_vs_11_586_2.pdf)

Algunenano:
Performance comparison for MVTs ([mvt_11_585_vs_mvt_11_586.pdf](https://github.com/CartoDB/Windshaft/files/2257054/mvt_11_585_vs_mvt_11_586.pdf)):
- No impact for points (expected).
- Lines that are not fully simplified (removed) are 2x slower.
- Polygons

Since the bug made geometries be simplified for a 256x256 tile instead of 4096x4096 the tolerance was 16 times more than it should. In reality, due to how Mapnik calculates tolerance the change means that for some zoom levels the tolerance is 10/100 times lower:
image.

The performance hit is something necessary once we fix the bug, but it also put us in a proper position to compare mapnik and pg-mvt renderers performance since now both should generate tiles with, at most, the expected simplification [width / extent].

Algunenano:
Hey, this should be ready to review now. I haven't see any performance impact from the changes in tilelive-mapnik (removing step and eventEmitter) and there is a expected impact on MVTs with lines and polygons using the mapnik renderer.

Once this is merge, I'll create another PR adding support for variable extent, which enables, among others, one tile optimization, or requesting more simplified tiles if overzooming isn't necessary.

dgaubert:
I'll be on vacations, please @oleurud take care of the CR.

lib/windshaft/renderers/pg-mvt/renderer.js
```diff
@@ -45,11 +45,10 @@ module.exports = class PostgresVectorRenderer {
     /// @param z tile zoom
     /// callback: will be called when done using nodejs protocol (err, data)
     getTile (z, x, y, callback) {
-        const emptyTileMsg = 'Tile does not exist';
         const subqueries = this._getLayerQueries({ z, x, y });
 
         if (subqueries.length === 0) {
-            return callback(new Error(emptyTileMsg));
+            return callback(null, new Buffer(0));
```
oleurud: Should we return `headers` (not sure about `stats`) making this case equal to the `no empty` one?

test/acceptance/mvt.js
```diff
@@ -500,9 +500,9 @@ function mvtExtractComponents(geometry) {
     });
 }
 
-// Compares (with a tolerance of +- 1) an array of points
+// Compares (with a tolerance of +- 2) an array of points
 function mvtPointArray_cmp(arr1, arr2) {
-    arr1 = arr1.filter(p1 => !arr2.find(p2 => Math.abs(p1.x - p2.x) <= 1 && Math.abs(p1.y - p2.y) <= 1));
+    arr1 = arr1.filter(p1 => !arr2.find(p2 => Math.abs(p1.x - p2.x) <= 1 && Math.abs(p1.y - p2.y) <= 2));
```
oleurud: Sorry, I will need more info to understand why in the `y` axis difference change

Algunenano: I separated all my changes into 2 PRs and this should have been in the second. I'll remove it from this one.

lib/windshaft/renderers/pg-mvt/renderer.js
Algunenano: Nice point. This should never happen as the mapconfig should be validated properly. I'll add a test and fix it as `pg-mvt` renderer is not checking it properly.

Algunenano: Turns out both are necessary. All layers should have `sql` but it might not be applied (for example if the layer is only enabled at certain zoom levels), so both checks are necessary.

oleurud: I think so too. My comment was more superficial or more thinking of the module usage. For example: the dev could think that if no error, the renderer always returns the header  `Content-Type: application/x-protobuf`

Algunenano:
Superseeded by https://github.com/CartoDB/Windshaft/pull/659

PR #647, opened 2018-08-06
Add vector_extent and vector_simplify_extent

Algunenano:
WIP: Requires rebase after https://github.com/CartoDB/Windshaft/pull/646 is merged

Adds the capability to setup the layer/tile extent for MVTs by adding 2 new options in MapConfig: `vector_extent` and `vector_simplify_extent`.

Even though the `pg-mvt` can modify the extent per tile request, the `mapnik` renderer doesn't, and that's why it is set during the renderer creation; this means that changing it requires re-instantiating the map.

It is required that all layers of a map have the same values. The defaults are:
- 4096 / 256 if non declared.
- vector_extent / vector_extent if only vector_extent is declared.
- 4096 / vector_simplify_extent if only vector_simplify_extent is declared.

References CartoDB/Windshaft-cartodb#843

### DIFF
--- a/doc/MapConfig-1.8.0.md
+++ b/doc/MapConfig-1.8.0.md
@@ -1,0 +1,359 @@
+# 1. Purpose
+
+This specification describes [MapConfig](MapConfig-specification.md) format version 1.6.0.
+
+
+# 2. File format
+
+Layergroup files use the JSON format as described in [RFC 4627](http://www.ietf.org/rfc/rfc4627.txt).
+
+```javascript
+{
+    // OPTIONAL
+    // Version of this spec to use for validation.
+    // Defaults to "1.0.0".
+    version: "1.7.0",
+
+    // OPTIONAL
+    // default map extent, in map projection
+    // (only webmercator supported at this version)
+    extent: [-20037508.5, -20037508.5, 20037508.5, 20037508.5],
+
+    // OPTIONAL
+    // Spatial reference identifier for the map
+    // Defaults to 3857
+    srid: 3857,
+
+    // OPTIONAL
+    // maxzoom to be renderer. From this zoom tiles will respond 404
+    // default: undefined (infinite)
+    maxzoom: 18,
+
+    // OPTIONAL
+    // minzoom to be renderer. From this zoom tiles will respond 404. Must be less than maxzoom
+    // default: 0
+    minzoom: 3,
+
+    // OPTIONAL
+    // Extra tolerance around the map (in pixels) used to ensure labels crossing tile boundaries are equally rendered
+    // in each tile (e.g. cut in each tile).
+    //
+    // To configure `buffer-size` per tile format:
+    //  - buffersize: {
+    //      'png': 64,
+    //      'grid.json': 64,
+    //      'mvt': 0
+    //    }
+    //
+    // Buffer-size also can be configured through `cartocss`, in these cases the priority to apply the right `buffer-size` is:
+    //  - Raster: cartocss >> map-config >> default
+    //  - Vector: map-config >> default
+    //
+    // default: undefined
+    buffersize: {
+        'png': 64,
+        'grid.json': 64,
+        'mvt': 0
+    },
+
+    // REQUIRED
+    // Array of layers defined in render order
+    // Different kind of layers supported are described below
+    layers: [
+        {
+            // OPTIONAL
+            // string, identifier for the layer
+            // Can be used to access layers by this id
+            // When `undefined` value is provided id will be chosen per layer, usually taking the form `layer{index}`
+            // Defaults to `undefined`
+            id: 'layer-name-id',
+
+            // REQUIRED
+            // string, sets layer type, can take 4 values:
+            //  - 'mapnik'  - rasterized tiles
+            //  - 'cartodb' - an alias for mapnik, for backward compatibility
+            //  - 'torque'  - render vector tiles in torque format (to be linked)
+            //  - 'http'    - load tiles over HTTP
+            //  - 'plain'   - color or background image url
+            type: 'mapnik',
+
+            // REQUIRED
+            // object, set different options for each layer type
+            options: {
+                // See different options by layer type bellow this
+                // NOTE: Options not defined in the different layers will be discarded
+            }
+        }
+    ]
+}
+```
+
+## 2.1 Mapnik layers options
+
+```javascript
+{
+    // REQUIRED
+    // string, SQL to be performed on user database to fetch the data to be rendered.
+    //
+    // It should select at least the columns specified in ``geom_column``,
+    // ``interactivity`` and  ``attributes`` configurations below.
+    //
+    // It can contain substitution tokens `!bbox!`, `!pixel_width!`, `!scale_denominator!`,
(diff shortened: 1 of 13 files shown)


lib/windshaft/renderers/mapnik/factory.js
```diff
@@ -404,3 +408,20 @@ function prepareQuery(userSql, geomColumnName, geomColumnType, options) {
...
+function setLayerExtent(grainstoreOptions, mmlOptions) {
+    const extents = _.uniq(mmlOptions.layer_extents);
+    const def = extents.indexOf(undefined);
+    if (def !== -1) {
+        extents[def] = 4096;
+    }
+    const extentsDefined = _.uniq(extents);
+
+    if (extentsDefined.length > 1) {
+        throw new Error("Multiple extent values in mapConfig (" + extents + ")");
```
rochoa: We should document that we just support one extent for all the layers.

doc/MapConfig-1.8.0.md
```diff
@@ -0,0 +1,359 @@
...
+# 2. File format
+
+Layergroup files use the JSON format as described in [RFC 4627](http://www.ietf.org/rfc/rfc4627.txt).
+
+```javascript
+{
+    // OPTIONAL
+    // Version of this spec to use for validation.
+    // Defaults to "1.0.0".
+    version: "1.7.0",
```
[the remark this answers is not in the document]
Algunenano: :ok_hand:

doc/MapConfig-1.8.0.md
```diff
@@ -0,0 +1,359 @@
...
+
+        // REQUIRED
+        // string list of columns returned by attributes service
+        columns: ['column1', 'column2']
+    }
+
+    // OPTIONAL
+    // Extent of the layer for MVTs. Must be the same for all layers
+    // Defaults to 4096
+    vector_layer_extent: 4096
```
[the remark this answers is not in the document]
Algunenano: I don't see `vector_extent` anywhere. I've used it's what is being used internally in `mapnik-vector-tile` but I don't mind changing it.

lib/windshaft/renderers/mapnik/factory.js
Algunenano: It's in the doc:
https://github.com/CartoDB/Windshaft/blob/89b99cd0ce7cc17a20e197c46ec222a68ada1e49/doc/MapConfig-1.8.0.md#L181

Should I clarify it more?

doc/MapConfig-1.8.0.md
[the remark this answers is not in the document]
rochoa: Hahaha. No, it doesn't exist anywhere. I meant that being an option within the layer, using `vector_layer_extent` looked like smurf naming.

lib/windshaft/renderers/mapnik/factory.js
rochoa: I missed that. It's OK. :)

Algunenano:
I've renamed `vector_layer_extent` into `layer_extent` and I've added an initial implementation for `vector_simplify_extent` with default to 256.
- For Mapnik it uses the configuration options to get the equivalent simplification.
- For St_AsMVT, for now, it uses an extra simplification step when needed (`ST_Simplify(ST_RemoveRepeatedPoints(${geomColumn}, ${tol}), ${tol}, true)` as geometry. This makes it compatible with Postgis 2.4, 2.5 and our current 2.4 fork. A function that does this as part of St_AsMVTGeom is likely to be added in the (near?) future to both our fork and Postgis 3.0 but how is still to be decided.

It is still pending adding some extra tests checking how both new MapConfig settings interact. I might need to increase tolerances around the MVT comparison functions since the result, specially when TWKB approximation is used, tend to differ more when `vector_extent` is much bigger than `vector_simplify_extent`; but it's probably not all that important.

Algunenano:
I've already been bitten by `vector_simplify_extent`'s default while doing some tests. I think it would be better to set it as `if vector_extent is defined, then use vector_extent, else 256`. This is compatible with the current buggy behaviour and would match my expectations when you only set `vector_extent`.

What do you think @rochoa ? Does it sound reasonable?

Algunenano:
This should be now ready for review. Keep in mind that sits on top of #646 (so it includes those changes too).

rochoa:
> What do you think @rochoa ? Does it sound reasonable?

I missed this. We talked off-line about it, though. I agree. 😃

lib/windshaft/renderers/renderer_factory.js
```diff
@@ -17,7 +17,7 @@ function RendererFactory(opts) {
     opts.mvt = opts.mvt || {};
     this.opts = opts;
 
-    this.mapnikRendererFactory = new MapnikRenderer.factory(opts.mapnik);
+    this.mapnikRendererFactory = new MapnikRenderer.factory(Object.assign(opts.mapnik, opts.mvt));
```
rochoa: I guess this is not critical here but beware of the side-effects of `Object.assign`.

Algunenano: This is no longer needed. I added it because I started adding the extent parameters as options, but since I ended up removing that I'll revert this change.

lib/windshaft/renderers/pg-mvt/renderer.js
```diff
@@ -45,13 +47,17 @@ module.exports = class PostgresVectorRenderer {
     /// @param z tile zoom
     /// callback: will be called when done using nodejs protocol (err, data)
     getTile (z, x, y, callback) {
-        const emptyTileMsg = 'Tile does not exist';
-        const subqueries = this._getLayerQueries({ z, x, y });
+        const headers = {
```
oleurud: :+1:

lib/windshaft/renderers/pg-mvt/factory.js
```diff
@@ -139,15 +215,31 @@ PgMvtFactory.prototype = {
...
+            this.options.vector_extent = extent || DEFAULT_EXTENT;
+
+            getSimplifyExtent(layers, extent, (err, simplify_extent) => {
+                if (err) {
+                    return callback(err);
+                }
+
+                this.options.vector_simplify_extent = simplify_extent || DEFAULT_SIMPLIFY_EXTENT;
+
+                const columnNamesPromises = layers.map(layer => {
```
enekid: FYI: Just a little shorter notation when the arrow function consists of only a return statement:

```javascript
const columnNamesPromises = layers.map(layer => getLayerColumns(psql, layer););
```

lib/windshaft/renderers/pg-mvt/renderer.js
```diff
@@ -25,18 +25,20 @@ function extractMVT(data) {
 /// A renderer for a given MapConfig layer
 ///
 module.exports = class PostgresVectorRenderer {
+
     constructor (layers, psql, attrs, options) {
```
enekid: default parameters:
```javascript
constructor (layers, psql, attrs, options = {}) {
```

Algunenano:
Superseeded by https://github.com/CartoDB/Windshaft/pull/659

PR #649, opened 2018-08-24
Remove step map validator

oleurud:


lib/windshaft/backends/map_validator.js
```diff
@@ -41,106 +38,72 @@ MapValidatorBackend.prototype.validate = function(mapConfigProvider, callback) {
...
-                }
+        if (!mapnikLayersIds.length) {
+            return this._validateRemaningFormatLayers(mapConfigProvider, mapConfig, params, callback);
+        }
 
-                var checkFailingMapnikTileLayerFns = [];
-                mapnikLayersIds.forEach(function (layerId) {
-                    checkFailingMapnikTileLayerFns.push(validateMapnikTile(layerId));
+        let allLayers; // if layer is undefined then it fetchs all layers
+        this.tryFetchTileOrGrid(mapConfigProvider, params, token, 'png', allLayers, errAllLayers => {
```
dgaubert: `allLayers` is always undefined

oleurud: Previously, the behavior of the first function in the `step` was:
```
              var validateAllMapnikTileLayers = validateMapnikTile();	
              validateAllMapnikTileLayers(next);
```
and the function
```
        function validateMapnikTile(layerId) {	
            return function(done) {	
                self.tryFetchTileOrGrid(mapConfigProvider, _.clone(params), token, 'png', layerId, done);	
            };	
        }
```
So, the parameter `layerId` is `undefined` and this is the way used to fetch all the layers.

I have added the variable and the comment in order to make visible this behavior (in the same way used by `validateVectorLayergroup` method)

oleurud: Do I have your blessings?

PR #652, opened 2018-08-24
Remove step factory

oleurud:


lib/windshaft/renderers/mapnik/factory.js
```diff
@@ -192,54 +188,54 @@ MapnikFactory.prototype.getRenderer = function (mapConfig, format, options, call
...
-                query: query,
-                protocol: isMvt ? 'bridge:' : 'mapnik:',
-                xml: xml,
-                strict: !!params.strict
-            };
-
-            if (isMvt) {
-                uri.gzip = false;
-            }
+    this._mmlStore.mml_builder(params, mmlBuilderOptions).toXML((err, xml) => {
```
dgaubert: `this._mmlStore.mml_builder(..)` can throw errors. Please prevent it from `uncaught exeptions`.

lib/windshaft/renderers/mapnik/factory.js
```diff
@@ -192,54 +188,54 @@ MapnikFactory.prototype.getRenderer = function (mapConfig, format, options, call
...
+            strict: !!params.strict
+        };
+
+        if (isMvt) {
+            uri.gzip = false;
         }
-    );
+
+        // hand off to tilelive to create a renderer
+        tilelive.load(uri, function (err, source) {
```
dgaubert: Not 100% sure, this method can throw errors as well.

lib/windshaft/renderers/mapnik/factory.js
oleurud: Thank you

lib/windshaft/renderers/mapnik/factory.js
dgaubert: did you check it?

oleurud: Yes, I forget to comment on what I see.

The `tilelive.load` could throw if the callback is not a function. But it does not apply here

oleurud: Do I have your blessings?

PR #653, opened 2018-08-24
Remove step image renderer

oleurud:


oleurud:
Your suspicions were correct: https://github.com/CartoDB/node-mapnik/blob/v3.6.2-carto/src/mapnik_image.cpp#L3410-L3431

lib/windshaft/renderers/plain/image_renderer.js
```diff
@@ -20,60 +20,32 @@ ImageRenderer.prototype.getTile = function(z, x, y, callback) {
...
-                    headers: {},
-                    stats: {},
-                    x: xPos,
-                    y: yPos,
-                    reencode: true
-                });
-
-                coveredWidth += (xPos < 0) ? (imageWidth + xPos) : imageWidth;
-                tX++;
+    try {
```
dgaubert: Please, add a comment to indicate that it's completely intentioned. When we migrate to Node.js 10 we will handle this situation in a better way:
```js
async getTile (...) { 
  [...]
  const image = mapnik.Image.fromBytes(buffer);
  const tiles = getTilesFromImage(image, ...);
}
```

PR #654, opened 2018-08-24
remove step renderer factory

oleurud:


lib/windshaft/renderers/renderer_factory.js
```diff
@@ -79,18 +76,18 @@ RendererFactory.prototype.genericMakeRenderer = function(factory, mapConfig, par
...
-            return callback(err, renderer);
-        }
-    );
+    try {
+        factory.getRenderer(mapConfig, format, options, function (err, renderer) {
+            if (err) {
+                return callback(err);
+            }
+
+            const rendererAdaptor = factory.getAdaptor(renderer, format, onTileErrorStrategy);
```
dgaubert: 😈 

MILLION DOLLAR QUESTION: in case of `factory.getAdaptor()` throws an error, do you think we are preventing `uncaught exception` here?

dgaubert: FYI: `factory.getAdaptor()` can throw errors.

oleurud: I thought `getAdaptor` does not throws. Of course, if it throws, the current try/catch doesn't catch the error.

Checking again slowly, most of the `getAdaptor` methods makes:
```
return new BaseAdaptor(renderer, format, onTileErrorStrategy);
```
And BaseAdaptor doesn't throw.

But the Mapnik factory could throw because it instances TileliveAdaptor (and it throws https://github.com/CartoDB/Windshaft/blob/master/lib/windshaft/renderers/mapnik/adaptor.js#L55)

I think it is better to change `TileliveAdaptor`. How do you see it?

oleurud: Thanks for the heads up! 

Probably, these changes become in the ugliest code ever written

lib/windshaft/renderers/renderer_factory.js
```diff
@@ -79,18 +76,22 @@ RendererFactory.prototype.genericMakeRenderer = function(factory, mapConfig, par
...
-        },
-        function makeAdaptor(err, renderer) {
-            assert.ifError(err);
-            return factory.getAdaptor(renderer, format, onTileErrorStrategy);
-        },
-        function returnCallback(err, renderer){
-            return callback(err, renderer);
-        }
-    );
+    try {
```
dgaubert: Please add a comment that we'll fix this when `async/await` lands.

lib/windshaft/renderers/renderer_factory.js
```diff
@@ -79,18 +76,24 @@ RendererFactory.prototype.genericMakeRenderer = function(factory, mapConfig, par
...
-            assert.ifError(err);
-            return factory.getAdaptor(renderer, format, onTileErrorStrategy);
-        },
-        function returnCallback(err, renderer){
-            return callback(err, renderer);
-        }
-    );
+
+    // waiting for async/await to refactor
+    try {
```
rochoa: This double try/catch block smells. 😞 

Isn't guaranteed to `factory.getRenderer` to not throw? If it's not guaranteed, that's the problem.

oleurud: Yes, I called it "the ugliest code ever written" in a previous comment.

Now, when every PR related to "remove step" are merged in this one https://github.com/CartoDB/Windshaft/pull/658, I think we could remove the first one. But I need to double check it again

rochoa: Hahaha. Fair enough.

PR #661, opened 2018-09-12
pg-mvt: Implement timeout in getTile

### LINKED ISSUE #577: Innefective vars in ST_AsMVT queries
When handling a query to be passed to `ST_AsMVT` we inject Mapnik tokens (bbox, scale-denominator, pixel_width, pixel_height) and also [three additional values](https://github.com/CartoDB/Windshaft/blob/3f1056391a7aae2c7b25d0557eff7be7f494d36f/lib/windshaft/renderers/pg-mvt/renderer.js#L49-L51): `var_zoom`, `var_x`, `var_y`.

This is ineffective because `SubstitutionToken` [only supports the mapnik tokens](https://github.com/CartoDB/Windshaft/blob/3f1056391a7aae2c7b25d0557eff7be7f494d36f/lib/windshaft/utils/substitution_tokens.js#L2-L5).

If the additional var variables are desirable we should add them to SubstitutionToken, otherwise remove them. I find them very convenient except that they can only be used with queries to be sent to ST_AsMVT (not Mapnik). 

@rafatower : I guess you introduced these variables, do you remember what was the intention?

cc @dgaubert

### LINKED ISSUE #624: Implement timeout for MVT PostGIS renderer
As we do with others renderers, in case of the database takes too long to respond we need to time out the underlying query and return a proper error.

Algunenano:
Since you need to open a new connection to close a running query I had to move the connection parameters to the renderer, so I decided to move `getLayerColumns` too and add a `initDB` function to avoid handling promises in the constructor.

To get the limits inside this renderer I've modified the factory to add `mapnik.mapnik.limits` to `mvt`, which is not pretty but needed for compatibility.

Closes #577 
Closes #624

### DIFF
--- a/lib/windshaft/renderers/pg-mvt/renderer.js
+++ b/lib/windshaft/renderers/pg-mvt/renderer.js
@@ -139,7 +139,7 @@ module.exports = class PostgresVectorRenderer {
         this.psql = new PSQL(this.dbParams, this.dbPoolParams);
     }
 
-    // Init DB and get layer column names.
+    // Retrieve the columns of all layers for later usage when requesting tiles
     getAllLayerColumns (callback) {
         const columnNamePromises = this.layers.map(layer => getLayerColumns(this.psql, layer));
         Promise.all(columnNamePromises)
--- a/lib/windshaft/renderers/pg-mvt/renderer.js
+++ b/lib/windshaft/renderers/pg-mvt/renderer.js
@@ -1,6 +1,7 @@
-const Timer = require('../../stats/timer');
 const debug = require('debug')('windshaft:renderer:pg_mvt');
+const PSQL = require('cartodb-psql');
 const SubstitutionTokens = require('../../utils/substitution_tokens');
+const Timer = require('../../stats/timer');
 
 function shouldUseLayer(layer, zoom) {
     const minZoom = Number.isFinite(layer.options.minzoom) ? layer.options.minzoom : 0;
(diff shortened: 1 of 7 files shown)


lib/windshaft/renderers/pg-mvt/renderer.js
```diff
@@ -39,45 +137,80 @@ module.exports = class PostgresVectorRenderer {
...
-        const timer = new Timer();
 
-        timer.start('query');
-        this.psql.query(query, function (err, data) {
-            timer.end('query');
-            if (err) {
-                debug("Error running pg-mvt query " + query + ": " + err);
-                if (err.message) {
-                    err.message = "PgMvtRenderer: " + err.message;
+        const pr_getTile = new Promise((resolve, reject) => {
```
dgaubert: What about extracting this promise to a new method of the class.

lib/windshaft/renderers/pg-mvt/renderer.js
```diff
@@ -39,45 +137,80 @@ module.exports = class PostgresVectorRenderer {
...
+        });
+    }
+
+    /// API: renders a tile with the Renderer configuration
+    /// @param x tile x coordinate
+    /// @param y tile y coordinate
+    /// @param z tile zoom
+    /// callback: will be called when done using nodejs protocol (err, data)
+    getTile (z, x, y, callback) {
+        return this._getTileTimeout(z, x, y, this.render_limit, callback);
```
dgaubert: Why not move the content of`_getTileTimeout` to `getTile`?

lib/windshaft/renderers/pg-mvt/renderer.js
Algunenano: That was my initial idea, then I saw that I needed the query string to be able to cancel and decide to merge it all together, but as I can see now I can have a function generating the promise that receives the query as a parameter.

Algunenano:
> As a general advice: try to use camelCase for naming variables.

I'll review it. I think it comes from switching languages and styles between projects :sweat_smile:

PR #666, opened 2018-10-16
Fix mvt layer columns

dgaubert:


rafatower:
> Do you have postgis@2.4 installed?

Yes, I do. The check for the particular version is a little bit picky.

rafatower:
The check of PG version is here: https://github.com/CartoDB/Windshaft/blob/3fe1d7fcdce57ced1e2839ca7d9fa5ecc38c5481/test/acceptance/mvt.js#L7

so it must be set in an env variable.

PR #668, opened 2018-10-18
Fix cache entry exception

dgaubert:


lib/windshaft/cache/cache_entry.js
```diff
@@ -1,3 +1,5 @@
+const debug = require('debug')('windshaft:cache-entry');
```
oleurud: Happy to see this :)

PR #684, opened 2019-03-27
Shameful hack to avoid uncaught exception

dgaubert:
Related https://github.com/CartoDB/cartodb-platform/issues/5389

Avoids the following `uncaught exception`
```
[2019-03-26 17:09:21.693] [ERROR] [default] - Uncaught exception: Error
    at /node_modules/queue-async/build/queue.js:39:41
    at cb (/node_modules/windshaft/lib/windshaft/renderers/blend/renderer.js:31:17)
    at /node_modules/windshaft/lib/windshaft/renderers/base_adaptor.js:19:28
    at /node_modules/windshaft/lib/windshaft/renderers/torque/png_renderer.js:103:21
[2019-03-26 17:09:21.694] [ERROR] console - Couldn't get marker-file http://com.cartodb.users-assets.production.s3.amazonaws.com/pin-maps/pins38.svg
```


lib/windshaft/renderers/blend/renderer.js
```diff
@@ -21,21 +21,39 @@ Renderer.prototype.getTile = function(z, x, y, callback) {
...
         tileQueue.defer(function (z, x, y, done) {
             var cb = function (err, buffer, headers, stats) {
-                if (err) {
-                    return done(err);
+                // FIXME: Under hidden circustances `done` migth throw an error
+                // when somenthing behind `getTile` isn't behaving as
+                // queue-async expects. For instance, when renderer (torque-png) is not able
+                // to download a marker (Couldn't get marker-file http://...svg)
+                // This should be fixed properly by stopping of using async-queue and follow
+                // the standards like Promises. Bedsides, we will need to review the error-handling in
```
Algunenano: `Bedsides` :bed:

dgaubert: :trollface:

lib/windshaft/renderers/blend/renderer.js
```diff
@@ -21,21 +21,39 @@ Renderer.prototype.getTile = function(z, x, y, callback) {
...
+                        buffer: buffer,
+                        headers: headers,
+                        stats: stats,
+                        x: 0,
+                        y: 0,
+                        reencode: true
+                    });
+                } catch (e) {
+                    // avoid to callback twice
+                    if (errored === false) {
```
Algunenano: Would it make more sense to use a different method instead of `forEach` to avoid this issue (calling the callback twice)?

https://stackoverflow.com/a/2641374 suggests using `some` or `every` but there might be a better alternative.

dgaubert: I wouldn't put much love on this hack. I prefer to leave it is as it is and let our `elves` to tackle it properly later. The more painful to our eyes the sooner we will fix it.

Algunenano: > The more painful to our eyes the sooner we will fix it.

LGTM. But I'm not conviced :smile:

oleurud: :laughing: ^ same here!

Another good option is to use the `for...of`

```
for (let renderer of this.renderers) { 
    .....
    } catch (e) {
         break;
    }
}

dgaubert: I cannot use any of your proposals as both of them requires synchronous processing. `.forEach` aims to prepare `tileQueue` for asynchronous processing. `.defer` enqueue callbacks to be performed afterward and calls `awaitAll` callback when it finishes.

oleurud: Ouch!

PR #686, opened 2019-04-09
Test against PG11 and Postgis 2.5

ethervoid:


ethervoid:
Test with PG11 are passing. When we should merge this to master? @Algunenano @javitonino ?

ethervoid:
retest this please

ethervoid:
retest this please

javitonino:
I think we should pass the PG version to the script as a parameter so we can do two runs in travis, one with PG10 and one with PG11, since we are going to be supporting both versions for some months.

ethervoid:
Passing using pg10 and pg11

dgaubert:
🚀

PR #694, opened 2019-06-25
[WIP] pg-mvt: Remove dependency on cartodb-postgresql extension

Algunenano:


Algunenano:
I'm reviewing Windshaft-cartodb with these changes, since the results of the aggregations depend on rounding :cry:

lib/windshaft/renderers/pg-mvt/renderer.js
```diff
@@ -268,7 +271,7 @@ module.exports = class PostgresVectorRenderer {
 
                 const stats = timer.getTimes();
 
-                return resolve([mvt || new Buffer(0), headers, stats]);
+                return resolve([mvt || new Buffer.alloc(0), headers, stats]);
```
dgaubert: ❤️

lib/windshaft/utils/cartodb_postgis_utils.js
```diff
@@ -8,22 +8,19 @@ function CartoDBPostgisUtils (tileSize, maxGeosize) {
 module.exports = CartoDBPostgisUtils;
 
 CartoDBPostgisUtils.prototype.cdbXYZResolution = function (z) {
-    var full_resolution = this.tileMaxGeosize / this.tileSize;
+    const full_resolution = this.tileMaxGeosize / this.tileSize;
     return full_resolution / Math.pow(2, z);
 };
 
 CartoDBPostgisUtils.prototype.cdbXYZExtent = function (x, y, z) {
```
dgaubert: Having in mind this module was only used by the old and deprecated `GeoJSON` module, we can modify it to our current needs.

dgaubert:
In `windshaft-cartodb` we still rely on CartoDB extension by using functions such as `CDB_XYZ_Resolution` and  `CDB_ZoomFromScale`.

dgaubert:
^^ here: https://github.com/CartoDB/Windshaft-cartodb/blob/master/lib/cartodb/models/aggregation/aggregation-query.js#L304

lib/windshaft/utils/cartodb_postgis_utils.js
```diff
@@ -1,29 +1,50 @@
...
 
+const Decimal = require('decimal.js');
+
 function CartoDBPostgisUtils (tileSize, maxGeosize) {
     this.tileSize = tileSize || 256;
-    this.tileMaxGeosize = maxGeosize || 40075017; // earth circumference in webmercator 3857
+    if (maxGeosize) {
+        this.tileMaxGeosize = new Decimal(maxGeosize);
+    } else {
+        // 6378137: earth circumference in webmercator 3857
```
rafatower: not to be picky, but isn't it the radius?

Algunenano: Yup, totally right

lib/windshaft/index.js
```diff
@@ -25,6 +25,7 @@ module.exports = {
     },
     backend: require('./backends'),
     version: version,
+    cartodb_utils: require('./utils/cartodb_postgis_utils'),
```
dgaubert: Let's try to use `CamelCase` at least for the public interface

dgaubert: what about naming it `utils`. Then in the client:

```js
const windshaft = require('windhsaft');
windshaft.utils.someMethod();
```

dgaubert: Nevermind, it's a class object. :(

dgaubert: Just thinking about the public interface:

```js
const windshaft = require('windhsaft');
const { TileHelper } = windshaft;
const tileSize = 256;
const tileHelper = new TileHelper(tileSize);
const resolution = tileHelper.getResolution({ z: 0 });
const extent = tileHelper.getExtent({ z: 0, x: 0, y: 0 });
```

thoughts?

Algunenano: The idea was to keep the same names / parameters to match the `cartodb` extension functions, but maybe changing to webMercatorUtils and providing getResoluton and getExtent as you say is clearer.

dgaubert: IIRC, we didn't use that module so we can rename it if we want to. `WebMercatorHelper` sounds great, no?

PR #696, opened 2019-07-12
Always run Postgis 2.4+ tests

Algunenano:
:cry:


dgaubert:
Don't cry, my friend. Nice to get rid of these conditionals

PR #698, opened 2019-07-16
ES6 goodies on Torque renderer

dgaubert:
- Use es6 class syntax
- Use template strings
- Stop using var, use const/let instead


Algunenano:
Do you plan to rebase this or should we close it?

PR #701, opened 2019-07-22
Remove underscore

dgaubert:


test/acceptance/cartocss-transformations.js
```diff
@@ -201,7 +199,11 @@ describe('cartocss transformation', function () {
             var mapConfig = TestClient.singleLayerMapConfig(scenario.sql, scenario.cartocss, scenario.cartocssVersion);
             var options = {
                 mapnik: {
-                    grainstore: _.defaults({ mapnik_version: scenario.mapnikVersion }, TestClient.grainstoreOptions)
+                    grainstore: Object.assign(
```
Algunenano: Why are you adding the extra `{}` and reordering here?

dgaubert: The reordering is because `_.defaults()` works in the other direction than `_.extend()` which has the same behavior than `Object.assign()`. I need to add an extra `{}` because the first parameter is the destination and we need to avoid overriding `TestClient.grainstoreOptions` as it's going to be used in other tests.

PR #705, opened 2019-09-10
[WIP] Use node-cartodb-psql with timeout

Algunenano:
Use the new parameter in PSQL.query to remove the need to have an extra event to handle the pg-mvt timeouts.

### DIFF
--- a/lib/windshaft/renderers/pg-mvt/renderer.js
+++ b/lib/windshaft/renderers/pg-mvt/renderer.js
@@ -39,33 +39,6 @@ function removeTrailingSemiColon(sql) {
     return sql.replace(/\s*;\s*$/, '');
 }
 
-
-function cancelQuery (query, dbParams, dbPoolParams, callback)
-{
-    let pg;
-    try {
-        pg = new PSQL(dbParams, dbPoolParams);
-    } catch (err) { return callback(); }
-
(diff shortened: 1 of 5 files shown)


lib/windshaft/renderers/pg-mvt/renderer.js
```diff
@@ -39,33 +39,6 @@ function removeTrailingSemiColon(sql) {
     return sql.replace(/\s*;\s*$/, '');
 }
 
-
-function cancelQuery (query, dbParams, dbPoolParams, callback)
```
dgaubert: 😍

lib/windshaft/renderers/pg-mvt/renderer.js
```diff
@@ -165,26 +138,20 @@ module.exports = class PostgresVectorRenderer {
     getTile (z, x, y, callback) {
         const subqueries = this._getLayerQueries({ z, x, y });
         if (subqueries.length === 0) {
-            return callback(null, new Buffer(0), { 'Content-Type': 'application/x-protobuf',
-                                                   'x-tilelive-contains-data' : false });
+            return callback(null, new Buffer.alloc(0), { 'Content-Type': 'application/x-protobuf',
```
dgaubert: 🤖

PR #708, opened 2019-10-24
Logging redis errors of MapStore

oleurud:
Related to https://github.com/CartoDB/support/issues/2177

The idea is to put this in production, gather stats and then remove it. Do you think I should do it with a more general-purpose?

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,3 +1,9 @@
+# Version 5.6.4
+2019-10-28
+
+Announcements:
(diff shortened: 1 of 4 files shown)


dgaubert:
Please, don't forget to update the NEWS.md before releasing

oleurud:
Finally, I am going to create a new `log4j` logger in `windshaft-carto`. The idea is to call it `windshaft` and it would write a JSON. I don't want to define the full logging thing, so I am going to focus only in this case, but it will be a good example for the full one.

PR #714, opened 2020-01-02
Replace istanbul by nyc as code coverage tool

dgaubert:


Algunenano:
jshint is complaning about I'm not sure what since `
/srv/test/results/compare-lib/js/coveringBad.js` doesn't make sense to me.

dgaubert:
🤔 
https://github.com/CartoDB/Windshaft/commit/53a4b315bcbd9b94bfd77ba631b3d7ae07edb42f

dgaubert:
I didn't use it before. I even did NOT aware of that. I guess we can remove it as it's a tool to see pixel diff when tests fail

Algunenano:
:fire: We can always recover it in the future if the need it (and remember it existed at some point).

dgaubert:
I'll do it in a separate PR, do you agree?

dgaubert:
https://github.com/CartoDB/Windshaft/issues/717

Algunenano:
All good

PR #716, opened 2020-01-03
Update README

dgaubert:


README.md
```diff
@@ -1,123 +1,124 @@
...
-- Redis >= 4
-- libcairo2-dev, libpango1.0-dev, libjpeg8-dev and libgif-dev for server side canvas support
+### PostGIS setup
 
-**Linux** Mapnik prebuilt module requires GLIBCXX_3.4.21 (GCC 5.1.0 / libstdc++-5 or higher). 
+A `template_postgis` database is expected. One can be set up with
 
-**OSX** It needs to be MacOS 10.9 or higher. It also requires pkg-config, you can install it with `brew install pkg-config`.
+```shell
+$ createdb --owner postgres --template template0 template_postgis
```
Algunenano: I think this is a renmant from the time when Postgis was installed as a script and not an extension. We should kill this and create and empty database and install postgis there instead.

PR #725, opened 2020-05-13
Node 12

dgaubert:


test/acceptance/torque-regressions-test.js
```diff
@@ -80,21 +80,18 @@ describe('torque regression', function () {
         var testClient = new TestClient(resolutionTwoMapConfig);
         testClient.getTile(13, 4255, 2765, { layer: 0, format: 'torque.json' }, function (err, torqueTile) {
             assert.ifError(err);
-            /* Sort torqueTile with x___uint8 ascending */
-            torqueTile.sort(function (a, b) {
-                return (a.x__uint8 > b.x__uint8);
-            });
+            torqueTile.sort((a, b) => a.x__uint8 - b.x__uint8 === 0 ? a.y__uint8 - b.y__uint8 : b.x__uint8 - a.x__uint8);
```
Algunenano, line 83: Ah, I see that you removed them and then discovered why they are there. Maybe we should leave a comment explaining that depending on the PG version, settings and so on the order of the output can change (and it's ok for it to change) so we need to sort it to compare it.

dgaubert: Yeah! I forgot why they were there 😅

dgaubert:
Added comments here: 1b5d2fefca2612a06e165a4d14c651dd49e2e58a

